PR #148, opened 2019-03-24
changelog: Remove broken issue links + fix diff links

kevinbader:


kevinbader:
FYI @Stezido @mathiashoeld

Issue #161, opened 2019-03-27
Ensure immutability of cache


gonzochic:
The cache is designed to be read-only and can only be updated through its own internals. But since the cache is also returning non-primitive datastructures (e.g. arrays, objects) it is possible to be updated unintentionally leading to undefined / undesired behaviour.

gonzochic:
Relates to #162

gonzochic:
Since #163, when querying the cache you will always receive a copy of the cached-state. This might be ressource intensive, because projects carry all the events inside the `log` array. We should think about separating the log from the rest of the data to make it only available if it is really needed.

PR #163, opened 2019-03-27
Cache immutability

### LINKED ISSUE #162: Apply functions update cache
Apply functions are intended to be used in the `*_eventsourcing.ts` files. In this files we are using `immer` to ensure that a apply function is not creating a side-effect, and instead returning the new value, instead of manipulating the original one.

The error occurs when we use the apply function outside of the `*_eventsourcing.ts`. For example during a write operation we often want to test if the test if the new event can be applied on the current state. Therefore we call apply inside the domain-logic. Additionally we are sometimes checking if the new state is different from the old state in order to reduce writes to the chain.

In this case using both of this features (apply + state comparison) we generate errorneous behaviour.

For example:
https://github.com/openkfw/TruBudget/blob/cc6c8fbf5ea1e067dd4e4a28e390ad53ea7c84c6/api/src/service/domain/workflow/subproject_permission_grant.ts#L56-L66

Where we call apply with the cached subproject calculation the new state for the subproject but also updating the cached subproject inside the apply function. Since the cache and the new state are now exactly the same the event is not added to the chain (wrong!) but the cache is updated (wrong!) giving the user the intention that the write succeeded because the change is visible on the UI, but in reality it isn't.

gonzochic:
Fixes #162 and #161 

The cache now always returns a copy of its data. This might reduce performance a bit, but ensures that nobody can edit the cache.

Additionally when the creation logic of the domain calls the "apply" function, it will always use a copy of the datasource.

### DIFF
--- a/api/src/service/domain/workflow/workflowitem_update.ts
+++ b/api/src/service/domain/workflow/workflowitem_update.ts
@@ -1,8 +1,5 @@
 import { produce as withCopy } from "immer";
-import Joi = require("joi");
-import isEqual = require("lodash.isequal");
 import { VError } from "verror";
-
 import { Ctx } from "../../../lib/ctx";
 import * as Result from "../../../result";
 import { BusinessEvent } from "../business_event";
@@ -17,6 +14,8 @@ import * as Project from "./project";
 import * as Subproject from "./subproject";
 import * as Workflowitem from "./workflowitem";
 import * as WorkflowitemUpdated from "./workflowitem_updated";
+import Joi = require("joi");
+import isEqual = require("lodash.isequal");
 
 export type RequestData = WorkflowitemUpdated.Modification;
 export const requestDataSchema = WorkflowitemUpdated.modificationSchema;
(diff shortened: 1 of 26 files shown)


api/src/service/domain/workflow/workflowitem_update.ts
```diff
@@ -61,25 +60,15 @@ export async function updateWorkflowitem(
...
-      if (Result.isErr(result)) {
-        throw new InvalidCommand(ctx, newEvent, [result]);
-      }
-
-      // Ignore the update if it doesn't change anything:
-      if (isEqual(workflowitem, result)) {
-        throw { newEvents: [], workflowitem };
-      }
+  // Check that the new event is indeed valid:
+  const result = withCopy(workflowitem, draft => WorkflowitemUpdated.apply(ctx, newEvent, draft));
```
kevinbader: if we use the name `produce` everywhere else, we should probably change the remaining `withCopy`s to `produce` as well..

gonzochic: Yes, we should discuss this. After reading the `immer` docs `produce` makes more sense for me + it matches the notation of the official doc.

gonzochic: renamed it in 8a4175b638f8a7a460cb47f7c1226e729bbcf5d8

PR #178, opened 2019-04-02
Review Readme and Docs

gonzochic:
The goal is to make the project as easy as possible for "new starters" to begin. This means, starting a  TruBudget Node and connecting a separate Slave Node to the master.


jzakotnik:
Super

PR #184, opened 2019-04-02
ui-fixes-for-beta-8

### LINKED ISSUE #141: Align columns in UI and prevent them from overlap
In GitLab by @lennart.ploen on Mar 18, 2019, 19:22

See screenshot attached

overlap-align

### LINKED ISSUE #159: ui: include digit grouping while typing in amount of Projected Budgets
Currently the separators are added after the input field loose focus only. An easy solution of providing this feature while typing (e.g. mui component) is preferred.

### LINKED ISSUE #171: ui: workflowitem cannot be edited when amount type change to allocated/disbursed
The reason for that behaviour is that the default exchange rate is not set when changing from N/A to allocated/disbursed

Stezido:
This PR resolves following ui-bugs:

Closes #141
Closes #159
Closes #171


mathiashoeld:
minor bug: workflowitem table: when the workflowitem title is too long, it reaches into the amount field. possible solution: introduce ellipses (...) to the title and display the full title in the info page

PR #188, opened 2019-04-03
Rely on docker-compose down to remove containers

kevinbader:


blockchain/startDevQuickly.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
```
phpai: remove this script

PR #195, opened 2019-04-05
API: Remove 'immer' dependency for projects

kevinbader:


kevinbader:
Closed in favor of #196

Issue #170, opened 2019-03-28
ui: authentication error shown after successful login


Stezido:
If a users login fails the next successful login will show an error message.
We should make sure that this error message is not shown after successful login.

PR #201, opened 2019-04-08
ui: Add new error code for failed login to saga

mathiashoeld:
When a user logs in with invalid credentials, the HTTP error code is now
400. Since it was 401 before, the snackbar after successful login did
not behave correctly.

Added error code 400 to snackbar message configuration.


mathiashoeld:
Closes #170

Issue #109, opened 2019-03-22
Show additional data/indicators on project level UI


phpai:
In GitLab by @lennart.ploen on Mar 7, 2019, 14:50

**Objective**: Add new data fields and indicators in the UI of the project level.

**Purpose**: Allow user to distinguish financing sources, access financial execution indicator and other project meta data attributes

**Concept**: 

The following data should be shown in the upper part of the UI on project level:
- Total budget
- Budget per source of finance (see issue 334)
- Proejcted budget
- Assigned budget
- Disbursed budget 

See attachments for details and definitions.

The following data should be made visible by pushing the meta data button (see issue 368 and 373) : project code, financing contract code(s), managing structure, action, activity, programme budgétaire.

For simplicity and in line with local applications, all amounts should be expressed in local currency. This is particularly relevant for co-financed projects with different currencies (e.g. € and local currency, a typical combination when the state co-finances part of the project) which need to be translated into one common currency in order to calculate the total. If the original amount is not the local currency, we propose that the amount is also displayed in the original currency.

[190318_Budgets.xlsx](https://github.com/openkfw/TruBudget/files/3002290/190318_Budgets.xlsx)

[190307_UI_projects.pptx](https://github.com/openkfw/TruBudget/files/3002291/190307_UI_projects.pptx)

phpai:
In GitLab by @lennart.ploen on Mar 18, 2019, 10:09

changed the description

phpai:
In GitLab by @lennart.ploen on Mar 18, 2019, 14:11

changed the description

phpai:
In GitLab by @kevinbader on Mar 22, 2019, 15:28

changed the description

kevinbader:
closed by #202

Issue #197, opened 2019-04-05
ui: project details do not show assigned budget


Stezido:
project view details is still summing up the amount field of all subprojects instead of projected budgets.

we need to refactor function `calculateMetrics()` in `ProjectDetails.js`

mathiashoeld:
possible overlap to #109

kevinbader:
closed by #202

PR #202, opened 2019-04-09
Adding analytics dashboard

gonzochic:
The charts on the project and subproject detail level should be moved to a new analytics component. The analytics component can do additional requests if necessary (only if it is opened).

I added an example implementation for the Subproject Details Page. It would be great to have similar adaption on the Project Details Page. Additionally we need to think about which type of useful charts we can provide (see #109).


Stezido:
The example looks good already. Additional thoughts of which useful charts we could provide:

- we only can provide projected budget analytics sorted by currency since sorting by organization is not possible because we do not have the information of organization saved in a workflowitem.
- we can not provide exact total budget analytics because we do not know the exchange rates of projected budgets on project/subproject level since they are variable.
- if the user provide the lacking information of exchange rates per currency manually the analytics can be provided

Stezido:
Closes #109 
Closes #197

PR #206, opened 2019-04-10
Sort swagger documentation

Stezido:
There are two suggestions according to the sort order:

1. sort by name
2. sort by usage (user/group, project/subproject/workflowitem)


Stezido:
Closes #207

PR #217, opened 2019-04-17
Exporting functionality

gonzochic:
Export all accessible data into an Excel-Sheet


gonzochic:
@kevinbader implemented your remarks. Left out pinging the readiness checks of the API's since I realized we have two of them. This leads us to the question: Is the export `Not Ready` if the "Test" Api is down but the "Prod" Api is up?

Feel free to merge :)

PR #243, opened 2019-04-26
Provisioning: Add Groups

### LINKED ISSUE #57: Provisioning: Add groups
In GitLab by @saschak_94 on Dec 19, 2018, 08:43

The current provisioning setup doesn't provision any groups. 
* [ ]  Clarify with Piet what the relevant groups are from business perspective
* [x]  Create a ```group.json``` file (similar to the users.json) in the provisioning project
* [x]  Implement group provisioning logic

mathiashoeld:
# Description
To test the functionality of groups on API level, groups are added to the provisioning of the blockchain. For this purpose, 2 groups are added: 
* Admins - Created with 2 users already in the API call
* Reviewers - Created without users, where 3 users are added and one is removed

Closes #57 

# How to test
Start an empty deployment and then run the provisioning. The groups should then be visible on the groups page of TruBudget (or listed when the API endpoint `/group.list` is called)


kevinbader:
@mathiashoeld in the issue there seems to be an open todo item:

> Clarify with Piet what the relevant groups are from business perspective

what about it?

mathiashoeld:
@kevinbader i'd suggest to create a separate issue for this. the groups can be edited afterwards.

PR #260, opened 2019-05-02
UI: bug fixes creation dialog

Stezido:
Previous merges cause issues on creation dialogs.

This PR fixes:
- responsive design workflowitem creation dialog is too small
- budget input field of dialogs is misaligned


frontend/src/pages/Common/CreationDialog.js
```diff
@@ -7,6 +7,14 @@ import DialogTitle from "@material-ui/core/DialogTitle";
 import isEmpty from "lodash/isEmpty";
 import strings from "../../localizeStrings";
 import CreationDialogStepper from "./CreationDialogStepper";
+import { withStyles } from "@material-ui/core";
```
gonzochic: Dont destructure in import since it is importing the whole "core" lib

Stezido:
```suggestion
import { withStyles } from "@material-ui/core/styles";
```

PR #277, opened 2019-05-09
Fix fetching of history items and bug fixes

### LINKED ISSUE #275: api returns duplicate history items due to limit argument
In case the offset is lower than the negative of the number of all history events, the offset is set to 0. That's expected, but apparently the limit is not decreased by the difference, which leads to duplicate history items being returned by the API.

### LINKED ISSUE #279: ui: Show correct text for projected_budget_updated and projected_budget_deleted in history
When projected budgets are edited or deleted, there is no text displayed, only the event type. 

This is solved by adding an entry in the `stringifyHistoryEvent` and corresponding entries in the language files.

mathiashoeld:
# What was changed
## Fetching of history items
The project/subproject/workflowitem history uses infinite scrolling.
Before this commit: When the last page is reached, the limit was not adjusted which lead to
overfetching.
Now: The limit is calculated based on the offsetIndex. If it is 0 (i.e.
the last page is reached), the limit is set to the modulus of the number
of events and the limit, which leaves only the events that were not
fetched before.
This also fixes the problem where the history drawer was populated with the wrong items after closing.

## Bugfixes
* Added stringify entries for update and delete projected budgets
* Fixed stringify of permission related history entries. The "formatPermission" method was restored to translate the permissions from the event to a value that can be read from the language files.

Closes #275 
Closes #279 

# How to test
Create more than 50 (current initial fetching limit) events for project/subproject/workflowitem, open the history and scroll to the bottom. The history items should be correctly loaded and no item should appear more than once. The number of items displayed should match the number of history events. (The number of history events can be checked via the `viewHistory`/`viewHistory.v2` endpoint of each resource.)
Each history entry should have a properly formatted text.

### DIFF
--- a/frontend/src/pages/Common/HistoryList.js
+++ b/frontend/src/pages/Common/HistoryList.js
@@ -5,6 +5,7 @@ import ListItemText from "@material-ui/core/ListItemText";
 import ListSubheader from "@material-ui/core/ListSubheader";
 import dayjs from "dayjs";
 import React from "react";
+import _isEmpty from "lodash/isEmpty";
 
 import { formatString } from "../../helper";
 import strings from "../../localizeStrings";
@@ -18,12 +19,16 @@ const styles = {
 
 export default function HistoryList({ events, nEventsTotal, hasMore, isLoading, getUserDisplayname }) {
   const eventItems = events.map((event, index) => {
-    const eventTime = event.getIn(["businessEvent", "time"]);
+    if (!(event.businessEvent && event.snapshot)) {
+      console.warn("The event does not have a business event or snapshot and will not be displayed", event);
+      return null;
+    }
+    const eventTime = event.businessEvent.time;
     return (
       <ListItem key={`${index}-${eventTime}`} className="history-item">
         <Avatar alt={"test"} src="/lego_avatar_female2.jpg" />
         <ListItemText
-          primary={stringifyHistoryEvent(event, getUserDisplayname)}
+          primary={stringifyHistoryEvent(event.businessEvent, event.snapshot, getUserDisplayname)}
           secondary={dayjs(eventTime).fromNow()}
         />
       </ListItem>
@@ -56,11 +61,12 @@ export default function HistoryList({ events, nEventsTotal, hasMore, isLoading, 
   );
 }
 
-function stringifyHistoryEvent(event, getUserDisplayname) {
-  const businessEvent = event.get("businessEvent");
-  const createdBy = getUserDisplayname(businessEvent.get("publisher"));
-  const eventType = businessEvent.get("type");
-  const displayName = event.getIn(["snapshot", "displayName"]) || "";
+const formatPermission = data => `"${strings.permissions[data.replace(/[.]/g, "_")]}"` || `"${data.intent}"`;
+
+function stringifyHistoryEvent(businessEvent, snapshot, getUserDisplayname) {
+  const createdBy = getUserDisplayname(businessEvent.publisher);
+  const eventType = businessEvent.type;
+  const displayName = snapshot.displayName || "";
 
   switch (eventType) {
     case "project_created":
@@ -72,7 +78,7 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
         strings.history.project_assign,
         createdBy,
         displayName,
-        getUserDisplayname(businessEvent.get("assignee"))
+        getUserDisplayname(businessEvent.assignee)
       );
     case "project_closed":
       return formatString(strings.history.project_close, createdBy, displayName);
@@ -80,19 +86,22 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
       return formatString(
         strings.history.project_grantPermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("grantee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.grantee),
         displayName
       );
     case "project_permission_revoked":
       return formatString(
         strings.history.project_revokePermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("revokee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.revokee),
         displayName
       );
-
+    case "project_projected_budget_updated":
+      return formatString(strings.history.project_projected_budget_updated, createdBy, businessEvent.organization);
+    case "project_projected_budget_deleted":
+      return formatString(strings.history.project_projected_budget_deleted, createdBy, businessEvent.organization);
     case "subproject_created":
       return formatString(strings.history.subproject_create, createdBy, displayName);
     case "subproject_assigned":
@@ -100,7 +109,7 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
         strings.history.subproject_assign,
         createdBy,
         displayName,
-        getUserDisplayname(businessEvent.get("assignee"))
+        getUserDisplayname(businessEvent.assignee)
       );
     case "subproject_updated":
       return formatString(strings.history.subproject_update, createdBy, displayName);
@@ -110,16 +119,16 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
       return formatString(
         strings.history.subproject_grantPermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("grantee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.grantee),
         displayName
       );
     case "subproject_permission_revoked":
       return formatString(
         strings.history.subproject_revokePermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("revokee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.revokee),
         displayName
       );
     case "workflowitems_reordered":
@@ -128,8 +137,8 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
     case "workflowitem_created":
       return formatString(strings.history.subproject_createWorkflowitem, createdBy, displayName);
     case "workflowitem_updated":
-      const update = businessEvent.get("update");
-      return update.has("documents") && !update.get("documents").isEmpty()
+      const update = businessEvent.update;
+      return update.documents && !_isEmpty(update.documents)
         ? formatString(strings.history.workflowitem_update_docs, createdBy, displayName)
         : formatString(strings.history.workflowitem_update, createdBy, displayName);
     case "workflowitem_assigned":
@@ -137,7 +146,7 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
         strings.history.workflowitem_assign,
         createdBy,
         displayName,
-        getUserDisplayname(businessEvent.get("assignee"))
+        getUserDisplayname(businessEvent.assignee)
       );
     case "workflowitem_closed":
       return formatString(strings.history.workflowitem_close, createdBy, displayName);
@@ -145,16 +154,16 @@ function stringifyHistoryEvent(event, getUserDisplayname) {
       return formatString(
         strings.history.workflowitem_grantPermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("grantee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.grantee),
         displayName
       );
     case "workflowitem_permission_revoked":
       return formatString(
         strings.history.workflowitem_revokePermission_details,
         createdBy,
-        strings.permissions[eventType],
-        getUserDisplayname(businessEvent.get("revokee")),
+        formatPermission(businessEvent.permission),
+        getUserDisplayname(businessEvent.revokee),
         displayName
       );
     default:
--- a/frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
+++ b/frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
@@ -1,21 +1,22 @@
 import React from "react";
 import { connect } from "react-redux";
+
+import { toJS } from "../../helper";
 import HistoryDrawer from "../Common/History/HistoryDrawer";
 import { hideHistory } from "../Notifications/actions";
-import { fetchProjectHistory } from "./actions";
+import { fetchNextProjectHistoryPage } from "./actions";
 
 function ProjectHistoryDrawer({
   projectId,
-  offset,
-  limit,
   doShow,
   events,
   nEventsTotal,
-  hasMore,
+  currentHistoryPage,
+  lastHistoryPage,
   isLoading,
   getUserDisplayname,
   hideHistory,
-  fetchProjectHistory
+  fetchNextProjectHistoryPage
 }) {
   return (
     <HistoryDrawer
@@ -23,8 +24,8 @@ function ProjectHistoryDrawer({
       onClose={hideHistory}
       events={events}
       nEventsTotal={nEventsTotal}
-      fetchNext={() => fetchProjectHistory(projectId, offset, limit)}
-      hasMore={hasMore}
+      fetchNext={() => fetchNextProjectHistoryPage(projectId)}
+      hasMore={currentHistoryPage < lastHistoryPage}
       isLoading={isLoading}
       getUserDisplayname={getUserDisplayname}
     />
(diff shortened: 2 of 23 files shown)


mathiashoeld:
@gonzochic the implementation has now been changed in the following way: 

- the history call is separated into two calls: fetch first items, fetch next items
- fetching the first items has a negative offset and gets e.g. the latest 50 elements, as well as the total number of history items at the point of opening the history for the selected resource
- from the total number of items, the next (positive) offset is calculated and the items are returned, but the total number of history items is **not** updated
- this means, that as long as the history drawer/tab is visible, the number of items does not get updated which prevents items from appearing twice or missing

frontend/src/pages/SubProjects/actions.js
```diff
@@ -101,9 +103,19 @@ export function setProjectHistoryOffset(offset) {
   };
 }
 
-export function fetchProjectHistory(projectId, offset, limit, showLoading = false) {
+export function fetchFirstProjectHistory(projectId, offset, limit, showLoading = false) {
```
gonzochic: Please find a better name e.g. `fetchMostRecentProjectHistoryItems`.And why do I need an `offset` here? For the user in his domain it is only interesting to choose the amount of history items, and the ID of the project.

frontend/src/pages/SubProjects/actions.js
```diff
@@ -101,9 +103,19 @@ export function setProjectHistoryOffset(offset) {
...
-    type: FETCH_PROJECT_HISTORY,
+    type: FETCH_FIRST_PROJECT_HISTORY,
+    projectId,
+    offset,
+    limit,
+    showLoading
+  };
+}
+
+export function fetchNextProjectHistory(projectId, offset, limit, showLoading = false) {
```
gonzochic: Please rename e.g. `fetchProjectHistoryItems(projectId, page)` and name the arguments according to the business domain. With this signature we could event drop the call to fetch the most-recent history

frontend/src/pages/SubProjects/actions.js
```diff
@@ -101,9 +103,19 @@ export function setProjectHistoryOffset(offset) {
   };
 }
 
-export function fetchProjectHistory(projectId, offset, limit, showLoading = false) {
+export function fetchFirstProjectHistory(projectId, offset, limit, showLoading = false) {
   return {
-    type: FETCH_PROJECT_HISTORY,
+    type: FETCH_FIRST_PROJECT_HISTORY,
+    projectId,
+    offset,
```
gonzochic: IMHO an action is business focused. Therefore I wouldn't expect to pass an `offset` and a `limit`. Especially not for the initial call.

frontend/src/pages/SubProjects/actions.js
```diff
@@ -101,9 +103,19 @@ export function setProjectHistoryOffset(offset) {
...
+    limit,
+    showLoading
+  };
+}
+
+export function fetchNextProjectHistory(projectId, offset, limit, showLoading = false) {
+  return {
+    type: FETCH_NEXT_PROJECT_HISTORY,
     projectId,
     offset,
```
gonzochic: Same as above. The redux action doesn't need to know how offset and limit works. What you ultimately want is the fetch the next `page`. Since we have a fixed `pageSize` we don't need to pass it. I would drop offset and limit completely and simply pass the `page` I want to fetch.

frontend/src/pages/WorkflowitemDetails/WorkflowHistoryTab.js
```diff
@@ -24,7 +26,14 @@ function WorkflowitemHistoryTab({
       hasMore={hasMore}
       isLoading={isLoading}
       getUserDisplayname={getUserDisplayname}
-      fetchNext={() => fetchWorkflowitemHistory(projectId, subprojectId, workflowitemId, offset, limit)}
+      // fetchNext={() => fetchFirstWorkflowitemHistory(projectId, subprojectId, workflowitemId, offset, limit)}
+      fetchNext={() => {
+        if (isFirstHistoryCall) {
```
gonzochic: If we only have one action creator (see above) we don't need to check anymore, wie simply pass the requested page (last page + 1)

frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
```diff
@@ -2,29 +2,28 @@ import React from "react";
...
   return (
     <HistoryDrawer
       doShow={doShow}
       onClose={hideHistory}
       events={events}
       nEventsTotal={nEventsTotal}
-      fetchNext={() => fetchNextProjectHistory(projectId, offset, limit)}
-      hasMore={hasMore}
+      fetchNext={() => fetchNextProjectHistoryPage(projectId)}
+      hasMore={currentHistoryPage !== lastHistoryPage}
```
gonzochic:
```suggestion
      hasMore={currentHistoryPage <== lastHistoryPage}
```

frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
```diff
@@ -33,20 +32,19 @@ function ProjectHistoryDrawer({
...
   };
 }
 
 const mapDispatchToProps = {
   hideHistory,
-  fetchNextProjectHistory
+  fetchNextProjectHistoryPage
 };
 
 export default connect(mapStateToProps, mapDispatchToProps)(ProjectHistoryDrawer);
```
gonzochic: a `toJS` is missing here in order to pass `events` as plan JS Object instead of passing an ImmutableJS object. Dumb components should not have ImmutableJS Objects: https://redux.js.org/recipes/using-immutablejs-with-redux#never-use-immutablejs-in-your-dumb-components

frontend/src/pages/SubProjects/SubProjectContainer.js
```diff
@@ -95,8 +94,7 @@ const mapDispatchToProps = (dispatch, ownProps) => {
     liveUpdate: projectId => dispatch(liveUpdateProject(projectId)),
     showSubprojectDialog: () => dispatch(showSubprojectDialog()),
 
-    openHistory: (projectId, offset, limit) => {
-      dispatch(fetchFirstProjectHistory(projectId, offset, limit, true));
+    openHistory: () => {
```
gonzochic:
```suggestion
    showHistory: () => {
```

frontend/src/pages/SubProjects/SubprojectHistoryDrawer.js
```diff
@@ -2,30 +2,29 @@ import React from "react";
...
   return (
     <HistoryDrawer
       doShow={doShow}
       onClose={hideHistory}
       events={events}
       nEventsTotal={nEventsTotal}
-      fetchNext={() => fetchNextSubprojectHistory(projectId, subprojectId, offset, limit)}
-      hasMore={hasMore}
+      fetchNext={() => fetchNextSubprojectHistoryPage(projectId, subprojectId)}
+      hasMore={currentHistoryPage !== lastHistoryPage}
```
gonzochic:
```suggestion
      hasMore={currentHistoryPage <== lastHistoryPage}
```

frontend/src/pages/SubProjects/SubprojectHistoryDrawer.js
```diff
@@ -34,20 +33,19 @@ function SubprojectHistoryDrawer({
 
 function mapStateToProps(state) {
   return {
-    offset: state.getIn(["workflow", "offset"]),
-    limit: state.getIn(["workflow", "limit"]),
     doShow: state.getIn(["workflow", "showHistory"]),
     events: state.getIn(["workflow", "historyItems"]),
```
gonzochic: Should not be an ImmutableJS object

frontend/src/pages/WorkflowitemDetails/WorkflowHistoryTab.js
```diff
@@ -2,59 +2,46 @@ import React from "react";
...
+  fetchNextWorkflowitemHistoryPage,
+  currentHistoryPage,
+  lastHistoryPage
 }) {
   return (
     <ScrollingHistory
       events={events}
       nEventsTotal={nEventsTotal}
-      hasMore={hasMore}
+      hasMore={currentHistoryPage !== lastHistoryPage}
```
gonzochic:
```suggestion
      hasMore={currentHistoryPage <== lastHistoryPage}
```

frontend/src/pages/WorkflowitemDetails/WorkflowHistoryTab.js
```diff
@@ -2,59 +2,46 @@ import React from "react";
...
       initialLoad={true}
     />
   );
 }
 
 function mapStateToProps(state) {
   return {
-    offset: state.getIn(["workflowitemDetails", "offset"]),
-    limit: state.getIn(["workflowitemDetails", "limit"]),
     events: state.getIn(["workflowitemDetails", "events"]),
```
gonzochic: We should not pass ImmutableJS Objects to dumb components

frontend/src/sagas.js
```diff
@@ -171,6 +168,43 @@ import { getExchangeRates } from "./getExchangeRates";
...
+    currentHistoryPage: state.getIn(["workflow", "currentHistoryPage"]),
+    lastHistoryPage: state.getIn(["workflow", "lastHistoryPage"]),
+    historyPageSize: state.getIn(["workflow", "historyPageSize"]),
+    totalHistoryItemCount: state.getIn(["workflow", "totalHistoryItemCount"])
+  };
+};
+const getWorkflowitemHistoryState = state => {
+  return {
+    currentHistoryPage: state.getIn(["workflowitemDetails", "currentHistoryPage"]),
+    lastHistoryPage: state.getIn(["workflowitemDetails", "lastHistoryPage"]),
```
gonzochic: Maybe we can remove this, since we are not using it in the sagas

frontend/src/sagas.js
```diff
@@ -794,110 +817,108 @@ export function* fetchAllProjectDetailsSaga({ projectId, showLoading }) {
...
+    let offset = 0;
+    if (totalHistoryItemCount === 0) {
+      // Before the first call, we don't know how many history items there are
+      // so we fetch a fixed number of the latest items
+      offset = -historyPageSize;
+    } else {
+      // After the first call we just fetch each page
+      // If the offset is below 0, we have reached the last page
+      // and can fetch the first events
+      offset = totalHistoryItemCount - (currentHistoryPage + 1) * historyPageSize;
```
gonzochic: This might be better:
```js
offset = Math.max(0, totalHistoryItemCount - (currentHistoryPage + 1) * historyPageSize);
```

frontend/src/sagas.js
```diff
@@ -794,110 +817,108 @@ export function* fetchAllProjectDetailsSaga({ projectId, showLoading }) {
...
-    });
-  }, showLoading);
-}
 
-export function* fetchNextProjectHistorySaga({ projectId, offset, limit, showLoading }) {
-  yield execute(function*() {
-    if (limit <= 0) {
-      return;
+    // If the offset is 0, we are on the last page and only need to fetch the remaining items
+    const limit = offset === 0 ? totalHistoryItemCount % historyPageSize : historyPageSize;
```
gonzochic: Do we really need to do this? I believe the API anyways only returns the amount of items it has

frontend/src/sagas.js
```diff
@@ -794,110 +817,108 @@ export function* fetchAllProjectDetailsSaga({ projectId, showLoading }) {
...
+
+    const { historyItemsCount, events } = yield callApi(api.viewProjectHistory, projectId, offset, limit);
+    const lastHistoryPage = historyPageSize !== 0 ? Math.ceil(historyItemsCount * 1.0 / (historyPageSize * 1.0)) : 1;
```
gonzochic: You can remove `1.0` since there is no `integer` type in Javascript

frontend/src/sagas.js
```diff
@@ -794,110 +817,108 @@ export function* fetchAllProjectDetailsSaga({ projectId, showLoading }) {
...
+    if (totalHistoryItemCount === 0 && historyItemsCount !== 0) {
```
gonzochic: Maybe it is easier to understand if we use a named variable here (e.g. `isFirstPage`)

frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
mathiashoeld: done

frontend/src/pages/SubProjects/SubprojectHistoryDrawer.js
mathiashoeld: done

frontend/src/pages/SubProjects/SubProjectContainer.js
mathiashoeld: renamed the action to `openHistory` since `showHistory` already exists as state variable

frontend/src/sagas.js
mathiashoeld: removed

frontend/src/sagas.js
mathiashoeld: done

frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
mathiashoeld: has to be `hasMore={currentHistoryPage < lastHistoryPage}`

frontend/src/sagas.js
mathiashoeld: `Math.max(0, totalHistoryItemCount - (historyPageSize * currentHistoryPage))` won't work because the limit has to be greater than 0 (restriction from the API). This means that we cannot use `limit = 0` for the case where we expect the API not to return anything.

I tested the case where the number of history items is a multiple of the page size (100 items and pagesize of 50) with this code 
```js
const limit =
      offset === 0 && totalHistoryItemCount % historyPageSize !== 0
        ? totalHistoryItemCount % historyPageSize
        : historyPageSize
```
and it works as expected. We could also use 
```js
    const limit =
      offset === 0 && totalHistoryItemCount % historyPageSize !== 0
        ? totalHistoryItemCount - historyPageSize * currentHistoryPage
        : historyPageSize;
```
which is basically the same.

frontend/src/sagas.js
mathiashoeld: done

frontend/src/sagas.js
mathiashoeld: done

frontend/src/pages/SubProjects/ProjectHistoryDrawer.js
mathiashoeld: changed view component, so it does not expect ImmutableJS objects anymore

frontend/src/pages/SubProjects/SubprojectHistoryDrawer.js
mathiashoeld: done

frontend/src/pages/WorkflowitemDetails/WorkflowHistoryTab.js
mathiashoeld: changed to `hasMore={currentHistoryPage < lastHistoryPage}`

frontend/src/pages/WorkflowitemDetails/WorkflowHistoryTab.js
mathiashoeld: was converted to pure JS objects

frontend/src/sagas.js
mathiashoeld: suggestion to make the code clearer and easier to read: 
```js
    const isLastPage = offset === 0;
    const remainingItems = totalHistoryItemCount - currentHistoryPage * historyPageSize;
    const limit = isLastPage && remainingItems !== 0 ? remainingItems : historyPageSize
```

frontend/src/pages/Common/HistoryList.js
```diff
@@ -18,7 +19,7 @@ const styles = {
 
 export default function HistoryList({ events, nEventsTotal, hasMore, isLoading, getUserDisplayname }) {
   const eventItems = events.map((event, index) => {
-    const eventTime = event.getIn(["businessEvent", "time"]);
+    const eventTime = event.businessEvent.time;
```
gonzochic: Please check if the event is valid by evaluating `businessEvent` and `snapshot`

frontend/src/sagas.js
```diff
@@ -830,16 +827,19 @@ export function* fetchNextProjectHistoryPageSaga({ projectId, showLoading }) {
...
     }
 
     // If the offset is 0, we are on the last page and only need to fetch the remaining items
-    const limit = offset === 0 ? totalHistoryItemCount % historyPageSize : historyPageSize;
+    const limit =
+      offset === 0 && totalHistoryItemCount % historyPageSize !== 0
+        ? totalHistoryItemCount % historyPageSize
+        : historyPageSize;
 
     const { historyItemsCount, events } = yield callApi(api.viewProjectHistory, projectId, offset, limit);
```
gonzochic: What happens if this call fails?

mathiashoeld: When setting the browser to "offline", the loading indicator ("Loading...") is displayed and no items are fetched.

frontend/src/pages/Common/HistoryList.js
mathiashoeld: done

PR #295, opened 2019-05-21
danger: Add check for missing e2e-tests

### LINKED ISSUE #256: Add dangerjs check for e2e tests for frontend code
The idea:

- if there is a change in `frontend/**`
- AND there is **no** change in `e2e-test/**`
- AND the title does **not** include "refactor"
- THEN warn about a missing end-to-end test for changed or added functionality.

Since there is a chance of false positives this should not cause the pipeline to fail.

Take a look at #254 to get an idea on how to implement this.

mathiashoeld:
Danger warns about missing E2E-tests for the following conditions:
- There are changes in the frontend
- AND there are no changes in the e2e tests
- AND the pull request is not marked as trivial

Closes #256

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -3,23 +3,42 @@ const { includes } = require("lodash");
 
 const apiSources = danger.git.fileMatch("api/src/**/*.ts");
 const blockchainSources = danger.git.fileMatch("blockchain/src/*");
-const frontendSources = danger.git.fileMatch("frontend/src/*");
+const frontendRootSources = danger.git.fileMatch("frontend/src/*.*");
+const frontendPageSources = danger.git.fileMatch("frontend/src/pages/*");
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -23,3 +23,8 @@ if (
 ) {
   warn("No CHANGELOG added.");
 }
+
+// If there are changes in the UI and PR is not marked as trivial, expect there to be E2E-test updates
+if (frontendSources.modified && !e2eTestSources.modified && !trivialPR) {
```
gonzochic: Asking myself how we should handle changes in the translations or tests. These changes don't require an update in the e2e tests

mathiashoeld: changed, so only changes in the `frontend/src` folder or `frontend/src/pages/` folder trigger the warning. 
also changed the message

PR #303, opened 2019-05-22
Fix notification issues

### LINKED ISSUE #288: ui: The last page of notifications displays too many items
Since the limit is fixed, we always fetch the same number of notifications for each page. On the last page however, only the remaining items should be displayed.

### LINKED ISSUE #292: ui: Wrong resource name displayed for some notifications
Case: User is assigned to a workflowitem. User does not have the permissions to see the workflowitem, but is allowed to see the parent subproject. The workflowitem is updated.

Current behaviour: The name of the workflow item's parent subproject is displayed in the notification page.

Expected behaviour: No name is displayed in the notification page, as it is the case for e.g. closing and assigning projects/subprojects.

### LINKED ISSUE #298: ui: Notification should contain Project / Subproject Name
On the notifications page, for every notification, it is not possibly to get any information about the related project. If possible we should add the at least the name. If you don't have the permissions we should add a remark about the missing permissions.

### LINKED ISSUE #299: ui: Assignee selections are overflowing
The assignee selections should draw an ellipsis if the name is too long.

### LINKED ISSUE #300: ui: Give different background color to unread notifications
To better distinguish unread notifications, they should have a slightly different background color than notifications that have been read.

### LINKED ISSUE #301: ui: Move 'Read All' button to the left
The 'Read All' button is currently on the right side of the notifications page, above the redirect buttons. This might lead to confusion because the icon indicating whether the message has been read is on the left side. 

Suggestion: Move the 'Read All' button to the left above the icon indicating whether the message has been read.

### LINKED ISSUE #302: ui: Improve 'view' button for notifications
The following things should be changed on the 'view' button of each notification on the notification page: 
- Use a smaller button, as it is a little too prominent on the page
- Don't surround the button with a blue circle
- Don't display the button if the user does not have the permissions to see the project/subproject/workflowitem

mathiashoeld:
# Changes

## Pagnination
The implementation for fetching notifications is changed. Instead of
keeping offset and limit in the state, the page is fetched by page
number instead (similar to the history).
The issue of overfetching on the last page was solved by calculating the
number of remaining items to be fetched for the last page.

Closes #288

## Prevent assignee selection from overflowing
When the name of the assignee is too long it gets shortened by ellipses
in the workflow item table.

Closes #299

## Display parent project and subproject name for notifications
The displaynames of the project and subproject are now displayed on the
notification page for each entry. If the user does not have permissions
to see the project/subproject it gets replaced by the word "Redacted".

Closes #298

## Display correct names in notifications
If the user does not have the permission to see a
project/subproject/workflowitem, the displayname is not displayed in the
notification and is instead left blank.

Closes #292

## Refactor Helpers
- Functions that are not used were removed
- Other functions were re-written for code clarity

## Move 'Read All' button
To increase usability, the 'Read All' button was moved to the left side
of the page. It is now above the icon that indicates whether the
notification has been read which makes its purpose clearer. The color
has also been changed to black instead of blue.

Closes #301

## Background color for unread notifications
To increase usability, the background color for unread notifications has
been changed to a light gray color.

Closes #300

## Remove view button if disabled
- The 'view' button has been changed to an icon button instead of
floating action button to make the look a little leaner
- The button has been surrounded by a tooltip which tells the user what
the button does
- If the user does not have the permissions to see the
project/subproject/workflowitem, the button is not displayed instead of
disabled

Closes #302

### DIFF
--- a/frontend/src/pages/Notifications/NotificationListItems.js
+++ b/frontend/src/pages/Notifications/NotificationListItems.js
@@ -1,6 +1,7 @@
 import { withStyles } from "@material-ui/core/styles";
+import IconButton from "@material-ui/core/IconButton";
+import Tooltip from "@material-ui/core/Tooltip";
 import Divider from "@material-ui/core/Divider";
-import Fab from "@material-ui/core/Fab";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemIcon from "@material-ui/core/ListItemIcon";
 import ListItemText from "@material-ui/core/ListItemText";
@@ -9,9 +10,10 @@ import Read from "@material-ui/icons/MailOutline";
 import LaunchIcon from "@material-ui/icons/ZoomIn";
 import dayjs from "dayjs";
 import React from "react";
+import classNames from "classnames";
 
-import { intentMapping, parseURI, isAllowedToSee } from "./helper";
-
+import { intentMapping, parseURI, getParentData, isAllowedToSee } from "./helper";
+import strings from "../../localizeStrings";
 const styles = theme => ({
   row: {
     display: "flex",
@@ -38,6 +40,9 @@ const styles = theme => ({
   unread: {
     flex: 1,
     opacity: 1
+  },
+  unreadMessage: {
+    backgroundColor: theme.palette.grey.main
   }
 });
 
@@ -59,12 +64,13 @@ const NotificationListItems = ({
       subprojectId: metadata.subproject ? metadata.subproject.id : undefined
     });
     const testLabel = `notification-${isRead ? "read" : "unread"}`;
+    const { projectDisplayName, subprojectDisplayName } = getParentData(notification);
     return (
       <div key={index}>
         <Divider />
         <ListItem
           component="div"
-          className={classes.row}
+          className={classNames(classes.row, !isRead ? classes.unreadMessage : false)}
           key={index}
           button={isRead ? false : true}
           data-test={testLabel}
@@ -73,8 +79,12 @@ const NotificationListItems = ({
           <div className={isRead ? classes.read : classes.unread}>
             <ListItemIcon>{isRead ? <Read /> : <Unread />}</ListItemIcon>
           </div>
-          <ListItemText className={classes.projectMetadata} component="div" primary={""} secondary={""} />
-
+          <ListItemText
+            className={classes.projectMetadata}
+            component="div"
+            primary={projectDisplayName}
+            secondary={subprojectDisplayName}
+          />
           <ListItemText
             data-test={`${testLabel}-${index}`}
             className={classes.title}
@@ -88,14 +98,15 @@ const NotificationListItems = ({
             secondary={createdAt}
           />
           <div className={classes.button}>
-            <Fab
-              size="small"
-              disabled={!isAllowedToSee(notification)}
-              color="primary"
-              onClick={() => history.push(redirectUri)}
-            >
-              <LaunchIcon />
-            </Fab>
+            {isAllowedToSee(notification) ? (
+              <Tooltip id="tooltip-inspect" title={strings.common.view}>
+                <div>
+                  <IconButton onClick={() => history.push(redirectUri)}>
+                    <LaunchIcon />
+                  </IconButton>
+                </div>
+              </Tooltip>
+            ) : null}
           </div>
         </ListItem>
       </div>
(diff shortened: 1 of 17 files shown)


frontend/src/pages/Notifications/NotificationListItems.js
```diff
@@ -41,6 +42,12 @@ const styles = theme => ({
   }
 });
 
+const styling = {
```
gonzochic: rename to `styles`

frontend/src/pages/Notifications/NotificationListItems.js
```diff
@@ -88,14 +101,15 @@ const NotificationListItems = ({
...
-              disabled={!isAllowedToSee(notification)}
-              color="primary"
-              onClick={() => history.push(redirectUri)}
-            >
-              <LaunchIcon />
-            </Fab>
+            {isAllowedToSee(notification) ? (
+              <Tooltip id="tooltip-inspect" title={strings.common.view}>
+                <div>
+                  <IconButton disabled={!isAllowedToSee(notification)} onClick={() => history.push(redirectUri)}>
```
gonzochic: We already checked if the user is allowed to see the notification (see Line 104)

frontend/src/pages/Notifications/actions.js
```diff
@@ -105,26 +103,18 @@ export function fetchHistoryItems(project, offset, limit) {
   };
 }
 
-export function markMultipleNotificationsAsRead(notificationIds, offset, limit) {
+export function markMultipleNotificationsAsRead(notificationIds, notificationPage) {
   return {
-    type: MARK_MULTIPLE_NOTIFICATION_AS_READ,
+    type: MARK_MULTIPLE_NOTIFICATIONS_AS_READ,
```
gonzochic: How do we know how big the page is? (25 or 50)?

gonzochic: and why do we need to send `page` and `ids`

frontend/src/pages/Notifications/helper.js
```diff
@@ -1,29 +1,91 @@
...
+    return notificationMetaData.subproject.displayName;
+  }
+  if (type === workflowitem) {
+    return notificationMetaData.workflowitem.displayName;
   }
   return "";
 }
 
+const isAllowedToSeeType = (notificationMetaData, type) => {
+  if (type === project) {
```
gonzochic: Make uppercase

gonzochic: Use switch case statement

frontend/src/pages/Notifications/helper.js
```diff
@@ -1,29 +1,91 @@
...
+    console.warn("Notification has no metadata");
+    return "";
+  }
+
+  const type = getType(notificationMetaData);
+  const displayName = getDisplayName(notificationMetaData, type);
+  const text = isAllowedToSeeType(notificationMetaData, type)
+    ? formatString(translation, displayName)
+    : formatString(translation, "");
+  return `${text} ${isAllowedToSeeType(notificationMetaData, type) ? "" : strings.notification.no_permissions}`;
```
gonzochic: You are evaluating the same condition twice

frontend/src/sagas.js
```diff
@@ -549,17 +557,27 @@ export function* getEnvironmentSaga() {
...
   yield execute(function*() {
-    // Get most recent items with negative offset
-    const { data } = yield callApi(api.fetchNotifications, 0 - offset - limit, limit);
+    const { data: notificationCountData } = yield callApi(api.fetchNotificationCounts);
+    const { notificationPageSize } = yield select(getNotificationState);
+
+    const totalNotificationCount = notificationCountData.total;
+
+    const numberOfNotificationPages =
+      notificationPageSize !== 0 ? Math.ceil(totalNotificationCount / notificationPageSize) : 1;
```
gonzochic: Not sure if this calculation is needed! We can always go to the first page when changing the page size

frontend/src/pages/Notifications/NotificationListItems.js
mathiashoeld: moved to styles and changed color in palette

frontend/src/pages/Notifications/NotificationListItems.js
mathiashoeld: removed since button is not displayed if user is not allowed to see the resource

frontend/src/pages/Notifications/helper.js
mathiashoeld: changed to uppercase and used switch statement

frontend/src/pages/Notifications/helper.js
mathiashoeld: moved addition of `strings.notification.no_permissions` into line 87

frontend/src/sagas.js
mathiashoeld: i think you mean the calculation in another file: https://github.com/openkfw/TruBudget/pull/303/files#diff-0d925ba8dfd401d545b4c938fb855ee6R39

this was removed and the first page is fetched instead.

frontend/src/pages/Notifications/actions.js
mathiashoeld: - the page size is saved in the state
- we send the page to reload after the messages have been marked as read
- we send the IDs because they are already available in the component. if we would send the page only, we would have to make another call to the API to get all the messages from the current page to extract the IDs, which would slow down the process.

PR #305, opened 2019-05-24
danger: Add check for new TODOs and console logs in API

### LINKED ISSUE #255: Add dangerjs check that warns about TODOs
A TODO could be there on purpose, so the pipeline should not fail. But both the author and the reviewer should be aware of it being added.

Take a look at #254 to get an idea on how to implement this.

mathiashoeld:
# Changed
## Two new warnings are added:
- If there are additional console logs in the API
- If there are new TODOs in any of the files

## Further changes:
- Created async block to execute async calls like
'structuredDiffForFile'
- Use Node 11 for Danger in order to use 'Array.prototype.flat()'
- Created functions to extract additions, deletions and normal code from
file changes

Closes #255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,15 +13,9 @@ stages:
     - "10"
   cache: npm
 
-.node-env-11: &node-env-11
-  language: node_js
-  node_js:
-    - "11"
-  cache: npm
-
 .dangerjs: &dangerjs
   stage: test
-  <<: *node-env-11
+  <<: *node-env
   script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,9 +13,15 @@ stages:
     - "10"
   cache: npm
 
+.node-env-11: &node-env-11
```
gonzochic: We should use the current LTS Versions (which is 10)

PR #306, opened 2019-05-24
ui: Don't display amount if no exchange rate is set

### LINKED ISSUE #297: Ui shows undefined exchnagerate if it is not set
Since the validation of a workflowitem is done once the worklfowitem is being closed, the exchangerate could be empty when the amounType is allocated or disbursed.
On frontend side we should handle empty exchangerate-fields instead of showing undefined:

image

mathiashoeld:
When a workflow item is updated via the API from amount type "N/A" to
"disbursed" or "allocated", the exchange rate does not need to be set
immediately (it is only needed if the workflowitem is closed). Therefore
it can happen that no exchange rate (or amount is set).
If no amount or exchange rate is set, the amount field is left blank.

Closes #297

### DIFF
--- a/frontend/src/pages/Workflows/WorkflowItem.js
+++ b/frontend/src/pages/Workflows/WorkflowItem.js
@@ -253,25 +253,29 @@ const isWorkflowSelectable = (currentWorkflowSelectable, workflowSortEnabled, st
 };
 
 const getAmountField = (amount, type, exchangeRate, sourceCurrency, targetCurrency) => {
-  let amountToShow = toAmountString(amount * exchangeRate, targetCurrency);
+  const amountToShow = toAmountString(amount * exchangeRate, targetCurrency);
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Workflows/WorkflowItem.js
```diff
@@ -253,25 +253,29 @@ const isWorkflowSelectable = (currentWorkflowSelectable, workflowSortEnabled, st
...
-      <div
-        style={{
-          paddingTop: "4px",
-          paddingLeft: "4px"
-        }}
-      >
-        {fromAmountString(exchangeRate) !== 1 ? amountExplaination : null}
-      </div>
+    <div>
+      {amount && exchangeRate ? (
```
gonzochic: Be aware that amount = 0 is false.

mathiashoeld: changed check to `amount !== undefined && exchangeRate !== undefined` and put it into separate variable `isAmountDisplayed`

Issue #79, opened 2019-03-22
Change password option


phpai:
In GitLab by @jzakotnik on Feb 1, 2019, 10:21

This is important for user management or e.g. changing password at first login

phpai:
In GitLab by @kevinbader on Feb 20, 2019, 11:52

assigned to @mathiashoeld

gonzochic:
Could be done together with #77

mathiashoeld:
Checklist: 

- [ ] New intent
- [ ] New event
- [ ] New domain & service + Test
- [ ] UI component & E2E test
- [ ] Documentation 
- [ ] Changelog

PR #309, opened 2019-06-03
api: Add endpoint to change a user's password

mathiashoeld:
# Changes

A new API endpoint was added that enables the change of a user's
password. For this, the following has been added:
* New intent 'user.changePassword'
* This new intent is given to every user per default so every user can change his/her own password
* New HTTP endpoint 'user.changePassword'
* New service
* New business event 'UserPasswordChanged'
* New unit test
* New business logic 'user_password_change'
* New event 'user_password_changed'

To be consistent with the rest of the API, the files for user handling
(creating, getting, sourcing) have been refactored.

Relates to #79 

# How to test
Call the API endpoint `user.changePassword` with `userId` and a new password and try authenticating the user using the old password (should not be possible) and then the new password (should be possible).

### DIFF
--- a/api/src/service/domain/organization/user_password_change.spec.ts
+++ b/api/src/service/domain/organization/user_password_change.spec.ts
@@ -1,0 +1,112 @@
+import { assert } from "chai";
+
+import { Ctx } from "../../../lib/ctx";
+import * as Result from "../../../result";
+import { hashPassword, isPasswordMatch } from "../../password";
+import { NotAuthorized } from "../errors/not_authorized";
+import { ServiceUser } from "../organization/service_user";
+import { newUserFromEvent } from "./user_eventsourcing";
+import { changeUserPassword, RequestData } from "./user_password_change";
+import { UserRecord } from "./user_record";
+
+const ctx: Ctx = { requestId: "", source: "test" };
+const root: ServiceUser = { id: "root", groups: [] };
+const alice: ServiceUser = { id: "alice", groups: ["alice_and_bob", "alice_and_bob_and_charlie"] };
+const bob: ServiceUser = { id: "bob", groups: ["alice_and_bob", "alice_and_bob_and_charlie"] };
+
+const dummy = "dummy";
+const passwordChangeUser: UserRecord = {
+  id: dummy,
+  createdAt: new Date().toISOString(),
+  displayName: dummy,
+  organization: dummy,
+  passwordHash: dummy,
+  address: dummy,
+  encryptedPrivKey: dummy,
+  permissions: {},
+  log: [],
+  additionalData: {},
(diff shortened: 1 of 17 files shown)


api/src/service/domain/organization/user_password_change.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      ...baseRepository,
+    });
+    assert.instanceOf(result, NotAuthorized);
+  });
+
+  it("The root user doesn't need permission to change a user's password", async () => {
+    const result = await changeUserPassword(ctx, root, requestData, {
+      ...baseRepository,
+    });
+    assert.isTrue(Result.isOk(result), (result as Error).message);
```
kevinbader: the operation has succeeded. But has it produced a new event?

api/src/service/domain/organization/user_password_change.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  it("A user can change another user's password if the correct permissions are given", async () => {
+    const result = await changeUserPassword(ctx, alice, requestData, {
+      ...baseRepository,
+      getUser: () =>
+        Promise.resolve({
+          ...passwordChangeUser,
+          permissions: { "user.changePassword": [alice.id] },
+        }),
+    });
+    assert.isTrue(Result.isOk(result), (result as Error).message);
```
kevinbader: the operation has succeeded. But has it produced a new event?

api/src/service/domain/organization/user_password_change.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        Promise.resolve({
+          ...passwordChangeUser,
+          permissions: { "user.changePassword": [alice.id] },
+        }),
+    });
+    if (Result.isErr(result)) {
+      throw result;
+    }
+
+    const sourcedUser = newUserFromEvent(ctx, passwordChangeUser, result[0]);
```
kevinbader: use the sourceUsers function here and pass the full list of events. or use reduce like this:

```typescript
const sourcedUser = result.reduce((user, event) => newUserFromEvent(ctx, user, event), passwordChangeUser);
```

api/src/service/domain/organization/user_password_change.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (Result.isErr(sourcedUser)) {
+      throw sourcedUser;
+    }
+    assert.isTrue(Result.isOk(result));
+    assert.isFalse(await isPasswordMatch(oldPassword, sourcedUser.passwordHash));
```
kevinbader: what about the new password, does it match?

PR #322, opened 2019-06-06
1.x.x release Portuguese language file updated

josenogueiradalmeida:


mathiashoeld:
thank you for the translations!

there seems to be a problem though as the npm test command fails. there are 201 additions and 208 deletions in the commit - can you make sure that all lines are preserved? there should be the same entries in all language files.

PR #324, opened 2019-06-11
Update portuguese.js

marcioonodera:
Revised portuguese translation - 20190611 - v2
(1st pull request ever ;)


mathiashoeld:
hi @marcioonodera! thanks for the translations!
there are just some minor issues with the spaces at the beginning of some lines. i will fix those, so you don't need to worry about it! :) 

just as information for the future: the lines in the translation file should all have the same indentation, so the start of each line is correctly aligned.

PR #332, opened 2019-06-19
ui: Only allow available budgets in subproject

### LINKED ISSUE #228: ui: On subproject level, only allow sources of finance that have projected budgets on project level
When creating a new subproject, the sources of finance can be entered freely. The source of finance field for the subproject should be selected from a drop down filled with sources of finance from the parent project.

mathiashoeld:
# Changes
When creating a subproject, the projected budget for a subproject can
only be selected from the projected budget of the parent project.
The organization is now a dropdown filled with the values of distinct
organizations in the project's projected budget.
E2E tests are added that test this functionality.

Closes #228

# How to test
Create a project with one or more projected budgets, then open the project and create a subproject - the organization field should be a drop down filled with the organizations from the projected budgets of the parent project.


frontend/src/pages/Common/Budget.js
```diff
@@ -1,19 +1,25 @@
...
-import TableBody from "@material-ui/core/TableBody";
-import TableCell from "@material-ui/core/TableCell";
-import TableHead from "@material-ui/core/TableHead";
-import TableRow from "@material-ui/core/TableRow";
-import TextField from "@material-ui/core/TextField";
-import DoneIcon from "@material-ui/icons/Check";
-import DeleteIcon from "@material-ui/icons/Delete";
-import EditIcon from "@material-ui/icons/Edit";
-import React from "react";
+import Button from '@material-ui/core/Button';
```
gonzochic: Shouldn't we use double quotes for imports?

frontend/src/pages/Common/Budget.js
```diff
@@ -35,6 +42,25 @@ export default class Budget extends React.Component {
...
+      ...new Set(
+        projectedBudgets.map(budget => {
+          return budget.organization;
+        })
+      )
+    ];
+
+    return distinctOrganizations.map((organization, index) => {
+      return (
+        <MenuItem key={`budget-${index}`} value={organization}>
```
gonzochic: dont use index, use the organization instead

PR #337, opened 2019-06-24
ui: Every user can see the user/groups table

### LINKED ISSUE #333: ui: Make user table available for every user
# Current situation 

The user needs to have one of the following permissions to see the table of users and groups: 
- global.createUser
- global.createGroup
- global.listPermissions

# Proposed change
Since the change password feature has been added (#325), the user should be able to update his/her password anytime. Therefore the table should be available to all users. The actions like changing the permissions or creating new users is checked on the table itself again, so no actions can be performed without the proper permission.

mathiashoeld:
# Changes 

Since every user can now change his/her password, every user should also
be able to see the users table. The check for the permissions global.createUser, global.createGroup and global.listPermissions have been removed.

E2E tests have been added to test this scenario. Additionally, the E2E
tests check for the availability of the buttons to add users and edit
user permissions.

Closes #333 

# How to test

Create a user with only basic permissions. This user should still see the "Users" menu in the navigation on the left and see the users table.

### DIFF
--- a/e2e-test/cypress/integration/users_spec.js
+++ b/e2e-test/cypress/integration/users_spec.js
@@ -1,30 +1,99 @@
+const testUser = {
+  userName: "dviolin",
+  password: "test"
+};
+
 describe("Users/Groups Dashboard", function() {
   before(() => {
     cy.login();
     cy.visit("/users");
   });
 
-  it("Show user dashboard", function() {
+  it("User dashboard is displayed for user with basic permissions", function() {
+    cy.login(testUser.userName, testUser.password);
+    cy.visit("/users");
+
+    // User dashboard should be visible to this user
     cy.location("pathname").should("eq", "/users");
     cy.get("[data-test=userdashboard]").should("be.visible");
   });
 
+  it("Button to add user not visible to user without proper permission", function() {
+    // Login as user without permissions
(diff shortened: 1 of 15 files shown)


e2e-test/cypress/integration/users_spec.js
```diff
@@ -4,27 +4,95 @@ describe("Users/Groups Dashboard", function() {
     cy.visit("/users");
   });
 
-  it("Show user dashboard", function() {
+  it("User dashboard is displayed ", function() {
+    // Create new user with only standard permissions
+    cy.addUser("another testuser", "anothertestuser", "test");
+
+    // Log in as new user
+    cy.login("anothertestuser", "test");
```
gonzochic: Is this a case to use fixtures?

e2e-test/cypress/integration/users_spec.js
```diff
@@ -4,27 +4,95 @@ describe("Users/Groups Dashboard", function() {
     cy.visit("/users");
   });
 
-  it("Show user dashboard", function() {
+  it("User dashboard is displayed ", function() {
+    // Create new user with only standard permissions
+    cy.addUser("another testuser", "anothertestuser", "test");
```
gonzochic: move to the before() function in order to make every test work independently

PR #350, opened 2019-07-17
ui: Change display behavior of workflow item buttons

### LINKED ISSUE #235: ui: Don't display edit and permissions button for workflow items if the user does not have proper permissions
As already done on project and subproject level, remove the buttons "edit" and "set permissions" on workflow item level, if the user does not have the permission to edit or set permissions or the workflow item is closed.

mathiashoeld:
When the user does not have the permissions to edit the workflow item,
permissions or status, the corresponding action button is not displayed.

Closes #235


mathiashoeld:
Use action button on all levels. 
Create new common component "ActionButton"

mathiashoeld:
@gonzochic please review

Issue #360, opened 2019-07-23
Not possible to add fields to resources and being backwards compatible


Stezido:
## Issue
The validation of Trubudget (Joi) does not allow additional fields and would respond with an validation error. Currently the api ignores the whole resource if an unknown field is found. This leads to not showing whole projects if they have an additional field of a new trubudget version. 

_Originally posted by @mathiashoeld in https://github.com/openkfw/TruBudget/pull/344#issuecomment-507724002_
Compatibility issue: When projects with tags are created in the new version, it is not available in earlier versions that do not have this issue (<1.3.0).
## Solution
Joi has an option to strip all additional unknown fields (stripUnknown).

Following table shows what would happen if a read or write intent is done with new fields and the stripUnknown option:

|   Intent   | Old Api version  | New Api version|
| ---------- | --------------------- | ------------- |
|     Read    | ignores new fields  | responds with new fields |
|     Write    | rejects request        | uses new fields  |

## Consequences
The disadvantage of stripping unknown fields is that there only will be an error if a required field is missing. So if a non required field is misspelled, the field is silently stripped. Note that api requests with misspelled fields are rejected by the fastify swagger schema and the user gets informed. The issue persists only internally, developer have to keep that in mind.   

## Affected Resources
Following resources could/should be expandable with additional fields:

- [ ] project
- [ ] subproject
- [ ] workflowitem
- [ ] user (user record)
- [ ] group
- [ ] notification
- [ ] global permissions

Stezido:
@gonzochic @kevinbader @mathiashoeld , what are your thoughts?

kevinbader:
agree - in order to support future versions, the validation _must_ ignore any additional fields. Unfortunately, I've never made sure the validation behaved as I thought it would - turns out it doesn't.

However, I'm not sure about that "Old API version rejects write request". For example, I don't see why creating a project in the old version shouldn't work. Any newer fields that might exist in future versions must be optional anyway - otherwise, the change wouldn't be backward compatible.

Stezido:
@kevinbader thanks for your participation. 
The old api version rejects write requests **with new fields** . As you've written creating a project does work on older versions in old format.

I will create a PR fixing that issue. The fix should be available in [v1.3.0](https://github.com/openkfw/TruBudget/milestone/19)

Stezido:
solved in #361

PR #361, opened 2019-07-24
add backwards compatibility for the future

Stezido:
## Description
The validation of Trubudget (Joi) does not allow additional fields and would respond with an validation error. Currently the api ignores the whole resource if an unknown field is found. This leads to not showing whole projects if they have an additional field of a new trubudget version.

The issue is described and discussed in #360 in detail.

## Implementation
Joi has an option to strip all additional unknown fields (stripUnknown).
This PR applies this option to all resources that must be sourced from stream items of the multichain:
-  project
-  subproject
-  workflowitem
-  user
-  group
-  notification
-  global permissions (global)
-  nodes

## Testing
To test backwards compatibility the function of our cache where stream items are parsed into business events could be unit tested. The endpoints must ignore new fields and may not respond with an validation error.
Testcases:
  - valid stream item - pass
  - future version stream item (additional field) - ignore
  - invalid stream item (a required field is missing in the stream item) - reject


Stezido:
@kevinbader could you review, please?

api/src/service/cache2.spec.ts
```diff
@@ -509,8 +515,401 @@ describe("The cache", () => {
...
+              time: "2019-07-25T12:03:25.315Z",
+            },
+          },
+          confirmations: 0,
+          blocktime: 1,
+          txid: "1",
+        },
+      ];
+      const addUnknownProperty = items => {
+        items.map(item => {
```
kevinbader: use forEach (map shouldn't be used with side effects, and the result of map should always be used)

api/src/service/cache2.spec.ts
```diff
@@ -509,8 +515,401 @@ describe("The cache", () => {
       assert.instanceOf(Result.unwrap_err(responseFromCache), NotFound);
     });
   });
+
+  describe("The BusinessEvent parsing", () => {
+    context("project events", async () => {
```
kevinbader: please describe the subject of this test - what is it testing?

api/src/service/cache2.spec.ts
```diff
@@ -509,8 +515,401 @@ describe("The cache", () => {
...
+      ];
+      const addUnknownProperty = items => {
+        items.map(item => {
+          item.data.json.project.additionalUnknownProperty = true;
+          return item;
+        });
+        return items;
+      };
+      const makeItemsInvalid = items => {
+        items.map(item => {
```
kevinbader: use forEach (map shouldn't be used with side effects, and the result of map should always be used)

api/src/service/cache2.spec.ts
```diff
@@ -509,8 +515,401 @@ describe("The cache", () => {
       assert.instanceOf(Result.unwrap_err(responseFromCache), NotFound);
     });
   });
+
+  describe("The BusinessEvent parsing", () => {
```
kevinbader: please add a comment somewhere here or at the top of the file that describes the general setup in a few words. if you read top-down, you really wonder where the assertions are and what this is testing in the first place.

api/src/service/cache2.spec.ts
Stezido: yes you're right. In this case I'm using the return value of map, I should right it like this instead:

`const addUnknownProperty = items => {`
`       return items.map(item => {`
`item.data.json.project.additionalUnknownProperty = true;`
`          return item;`
`        });`
`      };`

kevinbader: yes and no :) yes, because it _works_, but no, because the closure still has a side effect on item. Even if you'd ignore the return value of `map`, `item.data.json.project` would still be changed.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Added search feature for overview page [#347](https://github.com/openkfw/TruBudget/issues/347)
+- Added backwards compatibility for future versions [#360](https://github.com/openkfw/TruBudget/issues/360)
```
kevinbader: should go to "Fixed" imho

PR #363, opened 2019-07-30
update danger-js

Stezido:
update danger-js


Stezido:
Tested danger.js and it still works after updating to newest version 9.1.3.

PR #373, opened 2019-08-20
Add security restrictions for passwords (v2)

### LINKED ISSUE #370: Add security guidelines for passwords
New passwords should: 
- Be at least 8 characters long
- Contain at least one letter
- Contain at least one number

mathiashoeld:
# Changes
If a user changes his/her password, the new password has to adhere to the following conditions: 
- Minimal length of 8 characters
- At least one letter
- At least one number

Closes #370

# How to test
Change a user's password - you should not be able to create one that does not adhere to the above rules.


e2e-test/cypress/integration/change_user_password_spec.js
```diff
@@ -1,114 +1,130 @@
...
-      .should("be.visible")
-      .contains("Password successfully changed");
-  });
-});
+const testUserName = "passwordChangeUser";
+const testUserNamePassword = "test1234";
+
+describe("Users/Groups Dashboard", function() {
+  before(() => {
+    cy.login("root", "root-secret");
```
Stezido: I would outsource the root-secret into a variable

mathiashoeld: this is part of a bigger problem with e2e tests and will be handled separately

PR #374, opened 2019-08-20
Prohibit creation of second root

### LINKED ISSUE #369: UI: New user named root can be added
Current behaviour: Logged in as root able to create user called "root"

Expected behaviour: There should be only one user called "root"

miamarina:
# Changes

* New user with id 'root' cannot be created
* api: 'root' as an user id is blacklisted
* ui : An 'Invalid username' error is shown when 'root' is submitted
* unit test for user creation is added
* e2e test for component is added

Closes #369 

# How to test

* Frontend: User creation with user id 'root' should return error
* Api: User creation via `global.createUser` with user id 'root' should return error


api/src/service/domain/organization/user_create.ts
```diff
@@ -66,8 +66,13 @@ export async function createUser(
     additionalData: data.additionalData || {},
   };
 
+  const unfinishedBusinessEvent = UserCreated.createEvent(source, publisher, eventTemplate);
```
Stezido: **unfinishedBusinessEvent** could here also be a valid create Event.
I would suggest to rename it to createEvent and delete line 98 to prevent executing createEvent two times.

Stezido:
nice work, merge master into PR and clean up history

PR #387, opened 2019-08-28
Rework editing of permissions

### LINKED ISSUE #245: ui: Rework editing of permissions
## Current situation
When editing permissions, each click performs an API action and grants/revokes permissions. This is not ideal for some reasons: 
* If an error was made, it is written permanently to the chain
* The scenario where a user grants permissions, but cannot revoke them is possible - which combined with the first issue can cause problems
* There is no way to restore the original state during edit (i.e. no "cancel" button)

## Proposed solution
Similar to the drawer in the [batch edit function for workflowitems](https://github.com/openkfw/TruBudget/blob/master/doc/wiki/User-Guide/Projects/Workflowitem.md#update-multiple-workflowitems), the permissions should be changed in the permissions window without any action on API level and only after click of "apply" the changes are persisted to the chain. 
Additionally, add a "cancel" button that simply cancels the process and takes the user back to the previous screen. 
This should be done for global permissions as well as permissions for projects/subprojects/workflowitems.

mathiashoeld:
# Changes
The user now has to submit the changes that were made to the permissions of a project, subproject or workflow item. 

Closes #245 

# How to test
Change the permissions for e.g. a project, click submit and the permission should have changed. Try the same and hit cancel instead and the permissions should be the same as before opening the dialog.

### DIFF
--- a/frontend/src/pages/Overview/reducer.js
+++ b/frontend/src/pages/Overview/reducer.js
@@ -20,7 +20,9 @@ import {
   HIDE_PROJECT_ADDITIONAL_DATA,
   SHOW_PROJECT_ADDITIONAL_DATA,
   ADD_PROJECT_TAG,
-  REMOVE_PROJECT_TAG
+  REMOVE_PROJECT_TAG,
+  ADD_TEMPORARY_PROJECT_PERMISSION,
+  REMOVE_TEMPORARY_PROJECT_PERMISSION
 } from "./actions";
 
 const defaultState = fromJS({
@@ -39,6 +41,7 @@ const defaultState = fromJS({
   },
   idForPermissions: "",
   permissions: {},
+  temporaryPermissions: {},
   permissionDialogShown: false,
   currentStep: 0,
   initialFetch: false,
@@ -76,7 +79,8 @@ export default function overviewReducer(state = defaultState, action) {
       return state.merge({
         idForPermissions: defaultState.get("id"),
         permissionDialogShown: false,
-        permissions: fromJS({})
+        permissions: fromJS({}),
+        temporaryPermissions: fromJS({})
       });
     case SHOW_PROJECT_ADDITIONAL_DATA:
       return state.merge({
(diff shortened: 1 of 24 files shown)


frontend/src/pages/Overview/reducer.js
```diff
@@ -76,7 +79,8 @@ export default function overviewReducer(state = defaultState, action) {
       return state.merge({
         idForPermissions: defaultState.get("id"),
         permissionDialogShown: false,
```
Stezido: why not use the defaultState here?

frontend/src/pages/SubProjects/reducer.js
```diff
@@ -90,6 +93,7 @@ export default function detailviewReducer(state = defaultState, action) {
     case SHOW_SUBPROJECT_PERMISSIONS:
       return state.merge({
         permissions: fromJS({}),
```
Stezido: Why not use the defaultState here?

frontend/src/pages/Users/Dialog.js
```diff
@@ -86,58 +94,42 @@ const Dialog = props => {
       };
       break;
     case "editUserPermissions":
+      console.log(temporaryGlobalPermissions);
```
Stezido: remove this

frontend/src/pages/Users/reducer.js
```diff
@@ -94,10 +96,8 @@ export default function userDashboardReducer(state = defaultState, action) {
         dashboardDialogShown: false,
         userToAdd: defaultState.get("userToAdd"),
         authenticationFailed: false,
-        userPassword: "",
-        newPassword: "",
-        newPasswordConfirmation: "",
-        newPasswordsMatch: true
+        globalPermissions: fromJS({}),
```
Stezido: why not use the defaultState here?

frontend/src/pages/Workflows/reducer.js
```diff
@@ -182,6 +185,7 @@ export default function detailviewReducer(state = defaultState, action) {
       return state.merge({
         workflowItemReference: action.wId,
         permissions: fromJS({}),
+        temporaryPermissions: fromJS({}),
```
Stezido: why not use the defaultState here?

frontend/src/pages/Overview/reducer.js
mathiashoeld: done

frontend/src/pages/SubProjects/reducer.js
mathiashoeld: done

frontend/src/pages/Users/Dialog.js
mathiashoeld: done

frontend/src/pages/Users/reducer.js
mathiashoeld: done

frontend/src/pages/Workflows/reducer.js
mathiashoeld: done

PR #415, opened 2019-12-05
api: add check if organization does exist

### LINKED ISSUE #358: api: User is created although an error occured
### Description
If a user should be created via api with an organization that doesn't exist the api response is an error but also creates the user.

### Expected Behavior
After an error the user shall not be created.

Stezido:
### Description
If a user shall be created via api with an organization that doesn't exist the api response is an error but also creates the user.

### Tasks

- [x] Add a precondition check for organization in createUser
- [x] Add unit test

Closes #358

### DIFF
--- a/api/src/organization/organization.ts
+++ b/api/src/organization/organization.ts
@@ -1,5 +1,4 @@
 import { VError } from "verror";
-
 import logger from "../lib/logger";
 import * as SymmetricCrypto from "../lib/symmetricCrypto";
 import { Organization, WalletAddress } from "../network/model/Nodes";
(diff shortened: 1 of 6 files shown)


api/src/organization/organization.ts
```diff
@@ -100,6 +100,22 @@ async function ensureOrganizationAddress(
...
+    .v2_readStreamItems(streamName, streamItem, 1)
+    .then(items => items.map(x => x.data.json))
+    .then(items => items.some(_ => true))
+    .catch(error => {
+      if (error.kind === "NotFound") return false;
+      throw error;
+    });
+}
+
 export async function getOrganizationAddress(
```
mathiashoeld: use `getOrganizationAddressItem` to check if stream exists

api/src/organization/organization.ts
```diff
@@ -100,6 +99,17 @@ async function ensureOrganizationAddress(
   return organizationAddress;
 }
 
+export async function organizationExists(
+  multichain: MultichainClient,
+  organization: Organization,
+): Promise<boolean> {
+  if (await getOrganizationAddressItem(multichain, organization)) {
```
mathiashoeld: someone who is not familiar with the codebase might be confused by this line. please add a descriptive comment to this line (and maybe put it into a separate variable, e.g. `const organizationHasItems = await getOrganizationAddressItem(multichain, organization) ? true : false` and then use this as condition

PR #468, opened 2020-04-14
Add document download

### LINKED ISSUE #481: Store documents off-chain

phpai:
Closes #481 

TODO:

- [x] add documentation


api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -146,22 +148,54 @@ export async function createWorkflowitem(
...
+        docToUpload,
+      );
+
+      // Check that the event is valid:
+      const result = WorkflowitemDocumentUploaded.createFrom(ctx, workflowitemEvent);
+      if (Result.isErr(result)) {
+        throw { newEvents: [], errors: [new InvalidCommand(ctx, workflowitemEvent, [result])] };
+      }
+      return workflowitemEvent;
+    });
```
mathiashoeld, line 178: for cases like these we should use the `Result` type here

frontend/src/api.js
```diff
@@ -388,6 +388,34 @@ class Api {
...
+        `/workflowitem.downloadDocument?projectId=${projectId}&subprojectId=${subprojectId}&workflowitemId=${workflowitemId}&documentId=${documentId}`,
+        { responseType: "blob" }
+      )
+      .then(response => {
+        const url = window.URL.createObjectURL(new Blob([response.data]));
+        const link = document.createElement("a");
+        link.href = url;
+        const disposition = response.headers["content-disposition"];
+        let filename;
+        if (disposition && disposition.indexOf("attachment") !== -1) {
```
mathiashoeld, line 404: please add a helpful comment here so the reader knows what is going on

frontend/src/api.js
```diff
@@ -388,6 +388,34 @@ class Api {
...
+          const filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
+          const matches = filenameRegex.exec(disposition);
+          if (matches != null && matches[1]) {
```
mathiashoeld, line 407: please add a descriptive comment

api/src/service/domain/workflow/workflowitem_update.ts
```diff
@@ -89,7 +113,46 @@ export async function updateWorkflowitem(
...
-  return { newEvents: [newEvent, ...notifications], workflowitem: result };
+  // handle new documents
+  try {
+    let newDocumentUploadedEvents: BusinessEvent[] = [];
+    if (newEvent.update.documents && newEvent.update.documents.length > 0) {
+      const { documents } = newEvent.update;
+
+      newDocumentUploadedEvents = documents.map((d, i) => {
+        const docToUpload: UploadedDocument = {
+          base64: modification.documents ? modification.documents[i].base64 : "",
```
mathiashoeld, line 124: why is the value for `base64` set to empty string if there is no document at all? should this not be an error?

api/src/service/domain/workflow/document.ts
```diff
@@ -1,37 +1,51 @@
...
 import Joi = require("joi");
+import uuid = require("uuid");
+import * as Result from "../../../result";
 
 export interface StoredDocument {
+  // more or less a description
   id: string;
   hash: string;
+  // globally unique id
+  documentId: string;
```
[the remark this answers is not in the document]
Stezido: I checked them and added comments so it's more understandable what the different ids are for.
**id** =  id per workflowitem (same id in different workflowitems allowed)
**documentId** =  unique id in general, used to find the right document stored offchain

api/src/service/domain/workflow/workflowitem_document_download.ts
```diff
@@ -0,0 +1,56 @@
...
+  documentId: string,
+  repository: Repository,
+): Promise<Result.Type<WorkflowitemDocument.UploadedDocument>> {
+  // check for permissions etc
+  const workflowitem = await repository.getWorkflowitem(workflowitemId);
+  if (Result.isErr(workflowitem)) {
+    return workflowitem;
+  }
+
+  if (user.id !== "root") {
```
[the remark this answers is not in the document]
Stezido: Fyi: Currently all documents are shared between all nodes in the blockchain network. In the future access to documents shall be handled by multichain itself. For now the workflowitem.view permission is used to access a document of a workflowitem. I agree we need to clarify that, for this PR I will remove the root access.

api/src/service/domain/workflow/workflowitem_update.ts
Stezido: the `newEvent` or the prop `documents` are built from the modification.documents but they may not include the base64 so the base64 value has to be get from the modification. There is no way  the documents in the modification can be empty. So the empty string acts like a placeholder. The check has to be made since typescript compiler warns that modification.documents could potentially be empty.

PR #434, opened 2020-02-13
Placeholder for empty fields

melissaSchmidt0302:
Closes #31

PR #476, opened 2020-05-06
Revert "Placeholder for empty fields"

Stezido:
Reverts openkfw/TruBudget#434


Stezido:
Instead of reverting the changes, the e2e-tests shall be fixed.

PR #478, opened 2020-05-06
Fix email readiness error

### LINKED ISSUE #477: ui: Prevent UI from checking readiness of email service on every navbar creation
### Description
Currently the email readiness check is done when the userprofile/navbar container is built. If the server doesn't exist an error occurs in ui console. 

### Solution
Catch the error, and call the readiness endpoint (liveness) once while login.

melissaSchmidt0302:
Closes #477

### DIFF
--- a/e2e-test/cypress/integration/email_service_spec.js
+++ b/e2e-test/cypress/integration/email_service_spec.js
@@ -1,8 +1,4 @@
 describe("Email Service test", function() {
-  before(() => {});
-
-  beforeEach(function() {});
-
   it("If the email server is not available, user cannot change email address", function() {
     cy.login();
     cy.visit("/projects/");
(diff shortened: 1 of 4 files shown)


Stezido:
@Reviewer No Changelog entry needed since there is no change for the end user

Stezido:
The NODE_ENV variable cannot be used in react because no other values than "production" or "development" are accepted. I added a custom variable EXPOSE_STORE which is also not forwarded by react-scripts/react-app-rewired because it is only accepting custom variables with a REACT_APP_ prefix. The new env var REACT_APP_EXPOSE_STORE is forwarded as expected and locally it is working fine. In the test deployment the builds still have to be adpated. (Adding new env var to frontend)

Currently the store is exposed if REACT_APP_EXPOSE_STORE  and it's no production build. The master branch deployment is running in production mode so the store is not exposed that means the e2e-tests would fail. Maybe we should find another solution for testing the email feature in the frontend.

PR #479, opened 2020-05-11
GDPR warning for creating a user 

### LINKED ISSUE #466: Show a GDPR-warning when creating a user
## Description

When a user id is not GDPR conform its permanently put onto the blockchain. When creating a user a warning has to be showed to prevent the user creating a user with personal data. 
This warning is only a quick solution, for a solution that doesn't rely on user's actions a more complex technical solution is needed 

## Thoughts
One possible solution would be to save an autogenerated id onto the blockchain and the rest of the user-profile to a database connected to the master node of an organization. The id on the blockchain is then linked to the local database. The id can be deactivated by adding a deactivate flag to the id on the blockchain. The rest of the user profile can normally be deleted.  The only shared data via blockchain would be the user id. A downside would be that Trubudget can't use the user profiles of other organizations. The only way to achieve that is that the data is available locally in the database.... For ideas **HOW** to achieve that more thoughts are needed. 

#### Ideas and thoughts welcome!

melissaSchmidt0302:
Closes #466


frontend/src/pages/Common/CreationDialog.js
```diff
@@ -100,6 +100,7 @@ const CreationDialog = props => {
       data-test="creation-dialog"
     >
       <DialogTitle> {title}</DialogTitle>
+      {props.gdprWarning}
```
Stezido, line 103: I suggest to rename this variable since GDPR is europe specific. Maybe dataPrivacyWarning

melissaSchmidt0302: this variable is not in use, I missed to delete this.

PR #484, opened 2020-05-14
Add Workflowitem types

### LINKED ISSUE #483: Payment workflow with workflowitems
### Description
Workflowitems shall be used as follows: If user x assigns user y, user y gets permissions automatically and user x' write/admin permissions are revoked. So only the assignee can change the workflowitem. If the assignee changes the assignee's permissions changes too. 
The assignment should happen after a short confirmation to prevent unintentionally assigments.

#### Typed Workflowitems

Workflowitems can have specific types when creating them. These types are described in the docs or in the creation dialog. Each type has spefic behavior when e.g. assigning a user or closing the wfitem. For this feature request a type like e.g. "restricted" shall be added which behave like described above.
A user can choose the type of a workflowitem via dropdown at workflowitem creation.
At first these types shall be implemented in frontend only.
For Trubudget 2.0 these types can be specified on api/blockchain level.

#### Implement as general behavior 
Every Workflowitem should behavior like described above. Trubudget should be as general as possible so it's not possible to implement this solution 

#### Other Solutions
...brainstorming needed

Stezido:
### Description

Workflowitems shall have specific types when creating them. Each type has specific business rules applied at api level. At first there should be 2 types "general" and "restricted".

#### API
- [x] workflowitem.create works
- [x] Grant/Revoke permissions when assigning workflowitem "restricted"

#### UI
- [x] Create user input field for type when creating workflow item
- [x] restricted workflow item works as expected
- [x] Provide user information

Closes #483


api/src/service/domain/workflow/workflowitem_assign.ts
```diff
@@ -48,15 +51,47 @@ export async function assignWorkflowitem(
...
   );
+
   if (Result.isErr(assignEvent)) {
     return new VError(assignEvent, "failed to create event");
   }
 
+  let permissionsGranted;
+  let permissionsRevoked;
+
+  if (workflowitem.type === "restricted") {
```
Stezido, line 62: check if requesting user is permitted to grant/revoke permissions
only grant/revoke write/admin permissions - view permissions should be untouched
pay attention to the order of events. e.g. The last event must be the user's revoke "workflowitem revoke permission".

api/src/service/cache2-backwards-compatibility.spec.ts
```diff
@@ -130,6 +130,36 @@ const workflowitemStreamItems: any[] = [
           documents: [],
           permissions: {},
           additionalData: {},
+          type: "general",
```
Stezido, line 133: Is it type or workflowitemType? I think one should be deleted to test one with and one without type

api/src/service/domain/workflow/workflowitem_close.ts
```diff
@@ -125,5 +126,11 @@ export async function closeWorkflowitem(
       ),
     );
 
-  return { newEvents: [closeEvent, ...notifications] };
+  const workflowitemTypeEvents = applyWorkflowitemType(closeEvent, ctx, closingUser, workflowitemToClose);
+
+  if (Result.isErr(workflowitemTypeEvents)) {
+    return new VError(workflowitemTypeEvents, "failed to apply workflowitem types");
```
Stezido, line 132:
```suggestion
    return new VError(workflowitemTypeEvents, "failed to apply workflowitem type");
```

api/src/service/domain/workflow/workflowitem_update.ts
```diff
@@ -89,7 +90,13 @@ export async function updateWorkflowitem(
       ),
     );
 
-  return { newEvents: [newEvent, ...notifications], workflowitem: result };
+  const workflowitemTypeEvents = applyWorkflowitemType(newEvent, ctx, issuer, workflowitem);
+
+  if (Result.isErr(workflowitemTypeEvents)) {
+    return new VError(workflowitemTypeEvents, "failed to apply workflowitem types");
```
Stezido, line 96:
```suggestion
    return new VError(workflowitemTypeEvents, "failed to apply workflowitem type");
```

api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -146,7 +151,13 @@ export async function createWorkflowitem(
     return { newEvents: [], errors: [new InvalidCommand(ctx, workflowitemCreated, [result])] };
   }
 
-  return { newEvents: [workflowitemCreated], errors: [] };
+  const workflowitemTypeEvents = applyWorkflowitemType(workflowitemCreated, ctx, creatingUser, result);
+
+  if (Result.isErr(workflowitemTypeEvents)) {
+    return new VError(workflowitemTypeEvents, "failed to apply workflowitem types");
```
Stezido, line 157:
```suggestion
    return new VError(workflowitemTypeEvents, "failed to apply workflowitem type");
```

api/src/service/cache2-backwards-compatibility.spec.ts
melissaSchmidt0302: right one should be without type, I already noticed, I made a mistake when I rebased

CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-<!-- ### Added -->
+### Added
+
+- Changed date format to DD.MM.YYYY [#480](https://github.com/openkfw/TruBudget/issues/480)
```
Stezido, line 12: delete this line

PR #492, opened 2020-06-05
Allow disabling/deactivating of users

### LINKED ISSUE #241: Allow disabling of user
Since users are saved on the blockchain, there is no way to remove them. The only way to "deactivate" the user is to remove all permissions. Currently, this needs to be done manually for all projects, subprojects and workflow items, which is cumbersome and prone to error. 

The suggestion therefore is: Create the possibility to automatically remove all permissions for a given user. 
On API level: 
Use the "user.authenticate" permission to check if a user is allowed to log in and revoke it to "deactivate" the user. 

On UI level: Create a button that triggers the action on API level.

daniel-arnauer:
Closes #241

### DIFF
--- a/api/src/authz/intents.ts
+++ b/api/src/authz/intents.ts
@@ -67,7 +67,6 @@ export const globalIntents: Intent[] = [
   "global.createProject",
   "global.createUser",
   "global.createGroup",
-  "user.authenticate",
   "network.registerNode",
   "network.list",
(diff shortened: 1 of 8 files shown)


daniel-arnauer:
Branch has to be renamed since pipeline does not accept CamelCase in the name. To do that, a new Branch and a new PR are created.

PR #523, opened 2020-07-23
Allow typing comma separator for projected budget

### LINKED ISSUE #517: User cannot type comma separator while editing a projected budget
### Description
If a projected budget is edited and the user is typing in the comma separator him-/herself an error occurs that the money pattern is wrong.

### Solution
Allow typing in comma separators for projected budget

### LINKED ISSUE #559: Projected budget is not in the correct format
Depending on the current language, the projected budget has to be formatted:
Example 1000:

- english:  1,000.00
- other language: 1.000,00

### LINKED ISSUE #563: UI: workflowitem budget input is unstable
when typing the  workflowitem budget, the number gets strangely formatted, thus the number changes. This is because of the function toAmountString() and fromAmountstring().

melissaSchmidt0302:
Closes #517 
Closes #559 
Closes #563


daniel-arnauer:
e2e-test is missing

api/src/service/domain/workflow/money.ts
```diff
@@ -1,7 +1,7 @@
 import Joi = require("joi");
 
 export type MoneyAmount = string;
-export const moneyAmountSchema = Joi.string().regex(/^-?\s?[0-9]+(\.[0-9]+)?$/, "amount of money");
+export const moneyAmountSchema = Joi.string().regex(/^-?\s?(?:\d+|\d{1,3}(?:,\d{3})+)(?:\.\d+)?$/, "amount of money");
```
daniel-arnauer, line 4: the language specific comma sign ("," or ".") should be used here

Stezido (commented):
The current solution formats the budget before adding it to the other projected budgets. If I type in the wrong format and add it to the budgets suddenly the budget is in the right format but also it is another value. This is an unpredictable behavior. 
I would have expected that the value stays the same or the program doesn't let me add the budget in that format - giving an input field error message. Do you agree?

daniel-arnauer:
So i will ad an regex to check the input value. If the value does not match the regex, an input field error is thrown.
The conditions for regex in english language:
 
-  after ",", there must follow 3 numbers. Can occur infinitely
-  after ".", there can be an optional amount of numbers (will be rounded to two decimals). 
-  If "," or "." is set after an ".", the number is invalid

Other languages:
Same as english, but "," and "." are swapped

frontend/src/helper.js
```diff
@@ -94,6 +94,11 @@ export const toAmountString = (amount, currency) => {
   return accounting.formatMoney(amount, getCurrencyFormat(currency));
 };
 
+export const testMoneyInput = amount => {
```
Stezido, line 97: Remove the comment and rename the function so it's clear what the function does

frontend/src/pages/Common/Budget.js
```diff
@@ -20,33 +20,43 @@ const styles = {
...
+}) => {
   return (
     <TableCell align="right" data-test="saved-projected-budget-amount">
       {isEditing && editIndex === currIndex ? (
         <TextField
           label={strings.common.projected_budget}
           value={budgetAmountEdit}
-          onChange={e => setBudgetAmountEdit(e.target.value)}
+          onChange={e => {
+            if (/^[0-9,.-]*$/.test(e.target.value)) {
```
Stezido, line 40: explain regex or create own variable to describe what's the condition

frontend/src/languages/french.js
```diff
@@ -5,7 +5,8 @@ const fr = {
       decimal: ",",
       thousand: ".",
       precision: 2
-    }
+    },
+    numberRegex: /^([0-9]{1,3}.([0-9]{3}.)*[0-9]{3}|[0-9]+)(,[0-9]+)?$/
```
Stezido, line 9: describe regex in a comment

Stezido (changes requested):

frontend/src/pages/Common/Budget.js
daniel-arnauer: added var numberSignsRegex

frontend/src/helper.js
daniel-arnauer: renamed to checkAmountMoneyPattern

frontend/src/languages/french.js
daniel-arnauer: comments:
    // numberRegex describes values with "." as decimal seperator (matches e.g. 1000; 1,000; 1000.00; 1,000.00)
or
    // numberRegex describes values with "," as decimal seperator (matches e.g. 1000; 1.000; 1000,00; 1.000,00)

frontend/package.json
```diff
@@ -24,6 +24,7 @@
     "array-move": "^2.2.0",
     "axios": "^0.19.0",
     "chart.js": "^2.9.3",
+    "classnames": "^2.2.6",
     "connected-react-router": "^6.6.1",
```
daniel-arnauer, line 28: before merging, I will remove this line

Stezido (approved):

PR #524, opened 2020-07-24
api:  Result-Refactoring

Stezido:
### Description
`Result` should be used in each endpoint's http layer. This way in case of an error a detailed VError can be returned by the api.

### DIFF
--- a/api/src/service/document_validation.ts
+++ b/api/src/service/document_validation.ts
@@ -1,4 +1,5 @@
 import * as crypto from "crypto";
+import * as Result from "../result";
 
 /**
  * Returns true if the given hash matches the given document.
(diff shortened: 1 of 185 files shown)


api/src/service/document_validation.ts
```diff
@@ -9,10 +10,13 @@ import * as crypto from "crypto";
...
-
-  return computedHash === expectedSHA256;
+): Promise<Result.Type<boolean>> {
+  try {
+    const hash = crypto.createHash("sha256");
+    hash.update(Buffer.from(documentBase64, "base64"));
+    const computedHash = hash.digest("hex");
+    return computedHash === expectedSHA256;
+  } catch (error) {
+    return error;
```
daniel-arnauer, line 20: return new VError

daniel-arnauer:
Replace `throw new VError` to `return new VError` everywhere except the HTTP Layer.
Replace `return new Error` to `return new VError` everywhere.
Replace `Promise.reject` with `Result.isErr(.....)` everywhere.
 
This [commit message](https://github.com/openkfw/TruBudget/pull/524/commits/232ded843ced80c448cc4e1e5c19fba85fcc044c) is too long.

PR #539, opened 2020-09-09
api: use result and return errors in api functions

### LINKED ISSUE #531: api: Use Result in createEvent()/eventsourcing
### Description
To ensure consistency on api level all functions should return a `Result`.

### TODO
Errors should only be thrown on http layer. In service or domain layer the errors should only be returned and maybe wrapped with `VError`.

- [ ] createEvent
- [ ] event sourcing
- [ ] every other function where an error is thrown

georgimld:
some domain layer functions (createEvent, event sourcing and others) now have a Result return type and return errors instead of throwing them
Closes #531

### DIFF
--- a/api/src/service/domain/workflowitem_types/restricted.ts
+++ b/api/src/service/domain/workflowitem_types/restricted.ts
@@ -1,3 +1,4 @@
+import { VError } from "verror";
 import { workflowitemIntents } from "../../../authz/intents";
 import { Ctx } from "../../../lib/ctx";
 import * as Result from "../../../result";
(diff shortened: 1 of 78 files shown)


api/src/service/domain/workflowitem_types/restricted.ts
```diff
@@ -44,36 +45,47 @@ const createPermissionEvents = (
...
+        return new VError(createEventResult, "failed to create permission revoke event");
+      }
+      permissionsRevoked.push(createEventResult);
+    }
+  }
   // The permission for the revokeIntent should be revoked at last
-  const revokeIntentPermission = permissionsRevoked.filter(event => event.permission === revokeIntent);
-  permissionsRevoked = permissionsRevoked.filter(event => event.permission !== revokeIntent);
-
+  // tslint:disable-next-line:max-line-length
```
Stezido: delete this comment

api/src/service/store.ts
```diff
@@ -145,7 +145,7 @@ async function ensureStreamExists(conn: ConnToken, ctx: Ctx, name: string, kind:
         // Code -705 means the stream already exists - that's fine.
         return;
       }
-      throw new VError(err, `could not create ${kind} stream "${name}"`);
+      return new VError(err, `could not create ${kind} stream "${name}"`);
```
Stezido, line 148: missing function return type

Stezido (changes requested):
Please check my changes in `restricted.js`

Stezido (approved):

PR #540, opened 2020-09-09
api: Add check for viewHistory permissions

### LINKED ISSUE #486: Permissions for history are not set properly
For viewing the history, the intent project.viewHistory, subproject.viewHistory or workflowitem.viewHistory should be used in frontend and api. Currently, to view the history, the api checks only for workflowitem.view, subproject.viewDetails/viewSummary or project.viewDetails/viewSummary.

some of the files to change:
frontend\src\permissions.js
api\src\service\domain\workflow\project_history_get.ts
api\src\service\domain\workflow\subproject_history_get.ts
api\src\service\domain\workflow\workflowitem_history_get.ts

georgimld:
### Description
For viewing the history, only the intents project.viewHistory, subproject.viewHistory or workflowitem.viewHistory should be used on api level. Currently, to view the history, the api only checks for permissions on workflowitem.view, subproject.viewDetails or project.viewDetails.

### Solution
To avoid a breaking change the api should allow the request with viewDetails or viewHistory permissions. In Trubudget 2.0.0 this should be changed to only allow the request with viewHistory permission -> #547 

Closes #486

### DIFF
--- a/api/src/service/domain/workflow/project_history_get.spec.ts
+++ b/api/src/service/domain/workflow/project_history_get.spec.ts
@@ -89,6 +89,30 @@ describe("get project history: authorization", () => {
     assert.isTrue(Result.isOk(result), (result as Error).message);
   });
 
+  it("With only viewDetails permissions, a user can still get a project's history.", async () => {
+    const modifiedProject: Project = {
+      ...baseProject,
+      permissions: { "project.viewDetails": ["alice"] },
+    };
+    const result = await getHistory(ctx, alice, projectId, {
+      ...baseRepository,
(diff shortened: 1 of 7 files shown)


Stezido (approved):

Stezido (commented):
This should be changed in Trubudget 2.0, so that only the viewHistory permission allows the viewHistory request.

Stezido (changes requested):
Create unit test that checks if the user can get the history with  only viewDetails / only viewHistory / or both permissions. Currently only both are checked

Stezido (approved):

PR #546, opened 2020-09-22
Add e2e-test for worklfowitem badge-dialog

### LINKED ISSUE #404: e2e: Write tests for workflowitem batch edit
### Description
There are no tests for the workflowitem batch edit feature. Other features could break it.

### Todo
Write e2e-tests:
- [x] Reordering of open workflowitems is done on submit
- [x] Reordering button is disabled when user has no permission or there are only closed workflowitems 
- [ ] "Reordering mode" is enabled but closed workflowitems can't  be reordered
- [ ] "Reordering mode" is enabled but workflowitems can't  be reordered if there are redacted ones in between
- [x] Batch edit drawer opens on selecting the first workflowitem in reordering mode
- [ ] Batch edit drawer views current permissions/assignee if they are the same
- [ ] Preview dialog opens on clicking the drawers update button
- [ ] Preview dialog shows right actions including revoking actions (because grant permissions via batch edit overwrites the permissions of a workflowitem)
- [x] Changes will be applied correctly
 
reference to #408

### LINKED ISSUE #408: ui: Preview and confirmation dialog should use the same executeactions saga
### Description
Preview and confirmation dialog are currently using different sagas with the same goal. -> execute actions. The preview dialog should use the executeActions method and the actions should be in the same format. The behavior of executing actions in the batch-edit feature (preview dialog) should be similar to the permissions dialog.

#### Description
Currently the bath-edit feature is not acting in a expecting way for the user. It overwrites any permission the user wants to edit. For example if 7 user have view permissions on the selected workflowitems the batch edit feature would first revoke all 7 permissions and then grant view permissions to all users selected. A user is expected to only affect the users he/she selected.

*Hint: It's not possible to show the permissions in the batch-edit-drawer because multiple workflowitems could have different ones* 

#### Implementation
Separate grant and revoke permissions by offering a two part toggle button (grant/revoke) in the drawer. This way a check can be implemented if a selected user has already granted/revoked a permission or not.

### TODO

- [ ] Make both using the same actions format
- [ ] Make both use the same saga method
- [ ] The saga method should use other sagas instead of own api calls
- [ ]  Implement two part toggle button and separate grant/revoke permissions
 
reference to #404

daniel-arnauer:
Add a test for #404

### DIFF
--- a/e2e-test/cypress/integration/workflowitem_batch_spec.js
+++ b/e2e-test/cypress/integration/workflowitem_batch_spec.js
@@ -1,0 +1,70 @@
+/* eslint-disable no-unused-vars */
+describe("Workflowitem batch test", function() {
+  let projectId,
+    subprojectId,
+    workflowitem1,
+    workflowitem2,
+    workflowitem3,
+    workflowitem4,
+    workflowitem5,
+    apiRoute,
+    baseUrl;
+
+  before(() => {
+    cy.login();
+    baseUrl = Cypress.env("API_BASE_URL") || `${Cypress.config("baseUrl")}/test`;
+    apiRoute = baseUrl.toLowerCase().includes("test") ? "/test/api" : "/api";
+    cy.createProject("p-subp-batch", "workflowitem batch test").then(({ id }) => {
+      projectId = id;
+      cy.createSubproject(projectId, "workflowitem batch test").then(({ id }) => {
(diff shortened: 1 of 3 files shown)


daniel-arnauer:
Close PR because the branch name consists of uppercase letters

Issue #513, opened 2020-07-06
bc: backup feature is not working as expected


Stezido:
### Description
Sometimes the exported backup cannot be restored due to an invalid hash. The hash of the multichain directory is created by the blockchain project when the backup endpoint is called. The hash is stored in the metadata.yml file inside the tar'ed backup file. When trying to restore the backup the hash of the backup multichain directory is calculated again and is compared to the directory hash saved in the metadata.yml file. This hash seems to be incorrect sometimes? 

The bc endpoints GET/POST /chain should be checked again.

georgimld:
This problem happens when the hash stored in the metadata.yml file during backup is not the same as the hash that is being calculated during restore, so it is recognised as an invalid trubudget backup. The problem could either be in the creation of the metadata file, when the hash is generated, or when the new hash is created, in the restore feature. I believe that the hash in the metadata.yml file is the one that is invalid, because some backup files can never be restored and always lead to this problem. I didn't find a connection between a certain blockchain instance and this issue, seeing as the feature sometimes works (most of the time) and sometimes doesn't, on the same blockchain, without it being restarted and/or altered in some way.

Stezido:
Implemented a workaround in  #548
The issue is postponed to Milestone 2.0.0 since the way we hash/zip the Chain is inconsistent.
If we change the hash/zip function, backups made with earlier Trubudget versions wouldn't be useable anymore so the change is a breaking one.

galethil:
Issue with hashing was resolved by replacing MD5 hashing with SHA256 hashing and skipping of `metadata.yml` file from hash. Implementation done in issue #693

PR #548, opened 2020-09-29
Fix bugs for backup and restore

### LINKED ISSUE #239: api: Backup and restore should be async
E.g. notification.list times out during creation of backup

georgimld:
Frontend Changes:
- at the frontend level, live updates for notification.list, project.viewDetails and subproject.viewDetails are paused while a backup is created or restored, because the api is not available during that time

Blockchain Changes:
- in the restore function, the rpc password will be checked so no backups with other rpc passwords will be allowed
- from now on, the name of the organisation will also be saved in the metadata.yml file when creating a new backup
- if a backup contains the organisation in the metadata file, it will also be checked when a user tries to restore a backup
- a new script was created as a workaround to issue #513, this issue should be fixed in TruBudget 2.0.0
- documentation on how to use the script was added

e2e-tests:
- test that a backup file can be downloaded
- test that an invalid backup cannot be restored 
- test that a valid backup with another organisation cannot be restored
- test that a valid backup can be restored

Closes #239


Stezido (changes requested):
1. The backup e2e-test is failing maybe it's a timing issue (wait for responses before go on)
2. delete the failed test screenshots before pushing your changes (failed_restore_error.jpeg)
3. The names of the backup files are not quite clear - my suggestions are:
backup_valid ->valid_backup
backup_wrong_config -> backup_orga_test (If the orga is the "wrong config")
clean_backup -> provisioned_chain_backup
4. Create a Changelog entry because it's now possible to work around an " valid but invalid" backup file

e2e-test/cypress/integration/backup_spec.js
```diff
@@ -0,0 +1,158 @@
+import "cypress-file-upload";
+let baseUrl, apiRoute;
+baseUrl = Cypress.env("API_BASE_URL") || `${Cypress.config("baseUrl")}/test`;
+apiRoute = baseUrl.toLowerCase().includes("test") ? "/test/api" : "/api";
+
+describe("Backup Feature", function() {
+  beforeEach(() => {
+    //the user must be logged out before each test
+    cy.url().then(url => {
+      cy.log(url);
```
Stezido, line 10: remove it

Stezido (changes requested):
Gj, after clean up commit messages the PR can be merged

Stezido (approved):

PR #550, opened 2020-10-01
ui: Add url validation to filter empty parameters

### LINKED ISSUE #549: frontend: Check each GET request for empty get parameter
### Description
When sending a GET request the frontend is building the url e.g.: 
```
let url = `/workflowitem.viewHistory?projectId=${projectId}&subprojectId=${subprojectId}&workflowitemId=${workflowitemId}&offset=${offset}&limit=${limit}`;
```
If one of the parameters is not set e.g. limit is not set the request looks like this:
```
let url = `/workflowitem.viewHistory?projectId=123&subprojectId=1234&workflowitemId=12345&offset=0&limit=`;
```
The api checks the limit value and validate it's a positive integer. In this case limit is undefined but present so the api responds with a response code `400` and following message: `"message": "if present, the query parameter `limit` must be a positive integer"`

### Solution
Validating on api level is fine and the response is correct. The frontend has to be aware of sending empty query parameters.
So before returning the url in api.js the url has to be checked for such empty query parameters which have to be removed.

georgimld:
- the validateUrl function filters parameters that are empty or undefined

Closes #549

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -89,7 +89,7 @@ class Api {
       newPassword
     });
 
-  listUserAssignments = userId => instance.get(`/global.listAssignments?userId=${userId}`);
+  listUserAssignments = userId => instance.get(validateUrl(`/global.listAssignments?userId=${userId}`));
 
   createGroup = (groupId, displayName, users) =>
     instance.post(`/global.createGroup`, {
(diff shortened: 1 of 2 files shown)


frontend/src/api.js
```diff
@@ -465,4 +467,11 @@ const hasAttachment = response => {
   return dispositionHeader && dispositionHeader.indexOf("attachment") !== -1;
 };
 
+const validateUrl = url => {
```
Stezido, line 470: The name `validateUrl` is misleading since it changes the given value. I suggest to change the name to `removeEmptyQueryParams`.
I also suggest to add a function doc comment here to describe that the value is changed instead of returning a copy.
Each regex should have an explanation comment so it's more clear what's going on.

Stezido (changes requested):

PR #552, opened 2020-10-05
UI: Fix email-service bug

### LINKED ISSUE #551: UI: email info dissappears in user profile after refresh
After refreshing the frontend in the browser, the email info dissappears after clicking the button `data-test="show-user-profile"`.
Possible issue: state from `readiness` route is overwritten after refresh

daniel-arnauer:
Closes #551


Stezido (changes requested):
If the email service is not available, any email request may not be send. Prevent the saga from sending those requests when the email servcie isn't available.

PR #580, opened 2020-10-22
Group permission bug fix

### LINKED ISSUE #556: UI: Button for removing users from a group should only be shown if the permission exists
When editing a group, a user can click on the button to remove a user even if the user doesn't have the "group.removeUser" permission, which results in an error

georgimld:
Closes #556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Security -->
 
-## [1.15.0] - 
+## [1.15.0] - 2020-10-22
 
 ### Added
 
 - Script for fixing invalid backup files, as a workaround to issue [#513] [#548](https://github.com/openkfw/TruBudget/pull/548)
 
+### Fixed
+
+- The edit button is now only visible to the user that created it [#556](https://github.com/openkfw/TruBudget/issues/556)
 
 ## [1.14.0] - 2020-10-07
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -19,12 +19,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
-## [1.15.0] - 
+## [1.15.0] - 2020-10-22
 
 ### Added
 
 - Script for fixing invalid backup files, as a workaround to issue [#513] [#548](https://github.com/openkfw/TruBudget/pull/548)
 
+### Fixed
+
+- A group can now only be edited by the user who created it, other users don't see the edit button [#556](https://github.com/openkfw/TruBudget/issues/556)
```
Stezido, line 30: `A group can now only be edited by the user who created it` - that's the same behaviour as before

e2e-test/cypress/integration/groups_spec.js
```diff
@@ -43,4 +43,26 @@ describe("User/Groups Dashboard", function() {
         expect($th.first()).to.have.text("TestGroup");
       });
   });
+
+  it("User who created the group should be able to edit it", function() {
+    cy.get("[data-test=edit-group-TestGroup]").should("be.visible");
+  });
+
+  it("Other users should not be able to edit the group", function() {
```
Stezido, line 51: "Other" users are users that have not created the group - the test description should be more clear

Stezido (changes requested):

Stezido (approved):

PR #584, opened 2020-10-28
Fix falsy permission fetching in Confirmation Dialog

### LINKED ISSUE #544: UI: Confirmation dialog persists after leaving site
When using the "back" or "forward" button in the browser top left, the confirmation dialog persists in other projects with the data from the previous project (found this bug in adding or editing subprojects)

Solution: reset the confirmation dialog in an WillUnmount method.

### LINKED ISSUE #562: Issue on capacity of the responsible user to close a workflow item
[Bug] When a user assignes another user as the responsible of a workflow item, the new responsible can't close the workflow item.

Error   message : "Permissions required - Make sure you have required list   Permissions for all resources involved" when closing workflow item,   although veiw rights and "close workflow item" were granted and   possible to click on closure button
--

Closing a workflow

daniel-arnauer:
If a user has permission to close a workflowitem BUT no permission to list permissions of the workflowitem (workflowitem.intent.listPermissions), the confirmation dialog throws an error. The reason is that the confirmation dialog fetches permission of workflowitem by default, which is not necessary for the closing a   workflowitem. This also applies to project, subproject

Closes #562 
Closes #544

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Removed -->
 
-<!-- ### Fixed -->
+### Fixed
 
+- Fixed a bug where the user needed list-permissions that are not necessary [#562](https://github.com/openkfw/TruBudget/issues/562)
+- Fixed a bug where the confirmation dialog persists after pressing the browser's back button [#544](https://github.com/openkfw/TruBudget/issues/544)
 <!-- ### Security -->
 
 ## [1.16.0] - 2020-11-12
--- a/frontend/src/pages/Confirmation/ConfirmationContainer.js
+++ b/frontend/src/pages/Confirmation/ConfirmationContainer.js
@@ -35,12 +35,15 @@ import { executeOriginalActions } from "./executeOriginalActions";
 
 class ConfirmationContainer extends React.Component {
   componentDidUpdate(prevProps) {
-    // Fetch Permissions if grant/revoke permission confirmation required
-    if (
+    if (this.props.location.pathname !== prevProps.location.pathname) {
+      this.props.cancelConfirmation();
+    }
+
+    const isConfirmationDialogTriggered =
       this.props.confirmationDialogOpen !== prevProps.confirmationDialogOpen &&
-      this.props.confirmationDialogOpen === true &&
-      this.props.originalActions.some(action => !_isEmpty(action.payload))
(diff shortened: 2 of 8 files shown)


frontend/src/pages/Confirmation/ConfirmationContainer.js
```diff
@@ -27,22 +27,40 @@ import {
...
 import { applyOriginalActions, createAdditionalActions } from "./createAdditionalActions";
 import { executeOriginalActions } from "./executeOriginalActions";
 
 class ConfirmationContainer extends React.Component {
   componentDidUpdate(prevProps) {
-    // Fetch Permissions if grant/revoke permission confirmation required
-    if (
+    if (this.props.location.pathname !== prevProps.location.pathname) {
+      this.props.cancelConfirmation();
+    }
```
daniel-arnauer, line 41: this fixes #544

Stezido (commented):
How can I recreate the issue/error?
How is it solved?
Please give more context information in the PR.

frontend/src/pages/Confirmation/ConfirmationContainer.js
```diff
@@ -27,22 +27,40 @@ import {
...
+
+    // On these actions (e.g project.close), a Permission fetch is not needed
+    const isFetchingNeeded = !this.props.originalActions.some(action => {
```
Stezido, line 44: we should move the logic to the action payload which is sent by the saga. Another option is to move the logic to the reducer - If doing so you generate a new location where code has to be changed when adding a new confirmation intent. So I suggest to use the saga to decide if a confirmation needs project/subporject/workflowitem permissions or not.

Stezido (changes requested):

CHANGELOG.md
```diff
@@ -15,8 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Removed -->
 
-<!-- ### Fixed -->
+### Fixed
 
+- Fixed a bug where the user needed list-permissions that are not necassary [#562](https://github.com/openkfw/TruBudget/issues/562)
```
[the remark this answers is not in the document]
daniel-arnauer: done

frontend/src/pages/Confirmation/ConfirmationContainer.js
daniel-arnauer: done

Stezido (approved):

Issue #570, opened 2020-10-20
[Improvement] text for the "add a document" function is too long


a-dossa:
Replace "Nom du document du workflow" by "Nom du document"
<img width="486" alt="Nom du document" src="https://user-images.githubusercontent.com/72857322/96563961-de738480-12c2-11eb-946b-3aa58abc5bae.PNG">

Stezido:
fixed in #585

Issue #569, opened 2020-10-20
[improvement] Creating a group :  select users from dropdown list when assigning permissions


a-dossa:
Users -> Groups -> Edit Group / Select users from dropdown list when assigning permissions.

We need the same functionnality as the one used to select the responsable of a workflow. 
--

Stezido:
fixed in #585

Issue #568, opened 2020-10-20
Allow the accents in the "tag" field


a-dossa:
Accent can't be used on the tagg field. However, they are needed for the french users as well as for other languages.
--

Stezido:
fixed in #585

PR #585, opened 2020-11-02
Small bugfixes

### LINKED ISSUE #583: ui: fix bug in change password dialog
When the root user wants to change the password of another user, the text above the new password input field says "Type in new password for root" and should say the name of the user, not 'root'

georgimld:
- Fixed a bug in the dialog for changing the password
- Updated a french translation and used it as a label instead of a placeholder
- Adjusted the project tags so that it allows accents and upper and lowercase letters, adjusted the e2e-tests
- Refactored the component for user selection (also changed it to a functional component, changed the name, updated the e2e-tests)

Closes #583, #568, #570, #569


Stezido (changes requested):
Please describe the PR a little more so the reviewer understands quickly what he/she should review. What's the problem/s and how you intented to solve it/them.

1. root password test 
is fine gj ;)
1. accents tag
is fine gj ;)
1. add document text too long
If this should be renamed in the future again to something longer, the same issue will appear. 
So the solution would be to remove the label and set the right text as helper text only. Label and helper text is not needed
1. show "group add" user list
Why don't you use the already existing user list from the permissions dialog? Doesn't it fit? I thought only minor changes had to be done. e.g. change class to a component, use hooks, and make the userlist more general/reuseable. Then the issue I found would be solved too: The user I have added to the group still appears in the list.

Stezido (approved):

PR #587, opened 2020-11-03
api: Configure ESlint

georgimld:
- added eslint plugins and airbnb config for typscript
- fixed eslintrc file (it had a semicolon that was causing an error)


SamuelPull:
Esteemed colleagues, I took the liberty of rebasing master and squashing commits in one. Per Stefan's suggestion, all linter violations are overridden.

galethil (approved):

PR #595, opened 2020-11-05
Excel language feature

### LINKED ISSUE #567: language of Excel export of TruBudget
Exports of information about projects in TruBudget on Excel are in english. 
As english litteracy in African French speaking countries is low, exports should be in French

georgimld:
Frontend: 
- the language the user is logged in with is sent from the frontend to the excel-export server as a query parameter in the get request

Excel Export: 
- implemented the 'localized-strings' module and created translation files for all languages used in TruBudget
- when the export request arrives, the endpoint is checked and if it contains a language, then the language is set (index.ts)
- the column and sheet names of the table are set using the correct language (excel.ts)

e2e-test: 
- The excel file is downloaded in the cypress fixtures folder, is checked for the appropriate language using cy.tasks and then deleted

Closes #567

### DIFF
--- a/e2e-test/cypress/plugins/index.js
+++ b/e2e-test/cypress/plugins/index.js
@@ -12,6 +12,9 @@
 // the project's config changing)
 
 const axios = require("axios");
+const path = require("path");
+const Excel = require("exceljs");
+const fs = require("fs");
 
 function reportsReadiness(baseUrl) {
   return axios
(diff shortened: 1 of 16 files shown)


e2e-test/cypress/plugins/index.js
```diff
@@ -67,8 +70,49 @@ async function awaitApiReady(baseUrl) {
...
+
+async function deleteFile(file) {
+  if (fs.existsSync(file)) {
+    fs.unlinkSync(file);
+    return true;
+  }
+  return false;
+}
+
+async function checkFileExists(file) {
```
Stezido, line 91: add a timeout variable that controls for how long the filesystem should be checked

Stezido (changes requested):
use query parameter to decide which language to download

Stezido (approved):

PR #601, opened 2020-11-06
unhandled error in request handler

### LINKED ISSUE #561: prov: The provisioning exits when api is not reachable - SP 3
### Description
If the provisioning is started without having started the api the prov exits. The prov should retry instead.
While prov the prov should not generate any errors which is currently not the case.
Also sometime the prov is getting an error during prov and disconnects - it waits for 20 secs and go on without any problem again.

### TODO
Reproduce the error where prov exits without retrying.
Reproduce the error/s while provisioning.
Check solution ;)

SamuelPull:
Closes #561 

There was an unhandled exception in `getOrganizationAddressItem` in `api/src/organization/organization.ts` - multichain is `undefined`;
Perhaps it would be good to handle it in upper function call, `authenticateRoot` could check if it's not passing `undefined` arguments to  functions it calls. 

To replicate the bug, one way is to temporarily replace  in `api/src/index.ts`
`  const db = Multichain.init(rpcSettings);`   with   `const db = {};` and launch provisioning.

Provisioning exited fast, so it was not possible to attach a debugger. Added a launch configuration for debugging the service.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -32,6 +32,22 @@
       "request": "attach",
       "name": "Attach by Process ID",
       "processId": "${command:PickProcess}"
+    },
+    {
+      "name": "Launch provisioning",
+      "type": "node",
+      "request": "launch",
+      "cwd": "${workspaceFolder}/provisioning",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": [
+        "run",
+        "debug"
+      ],
+      "env": {
+        "NODE_ENV": "development",
+        "API_PORT": "8080",
+        "ORGANIZATION": "ACMECoro"
+      }
     }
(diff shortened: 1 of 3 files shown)


daniel-arnauer:
Please reset following files because they do not have actual changes (only format changes):
provisioning/src/index.js 
provisioning/src/lib.js

SamuelPull:
> Please reset following files because they do not have actual changes (only format changes):
> provisioning/src/index.js
> provisioning/src/lib.js
> 
> Rebase commits
> 
> If the Issue has been resolved, use `Closes #561 ` in the description

Hi @daniel-arnauer, files reset and commits squashed.

Stezido (approved):

PR #609, opened 2020-11-12
UI: scroll back to top button added

### LINKED ISSUE #579: [Low criticity] Easing the acess to the header of the main screen
As more and more project will be integrated in Trubudget, the list of project users will see on a the first screen of TruBudget will increase.
As a consequence, users will scroll down more and more after time to find their projects and will have to scroll up every time they want to go back to the header.

In other to avoid this, we propose to give a way to have an easy access to the header area (see image attached)
 as you scroll down.

Two wayS identified to do so : 

1. Add a button "go back at the top of the screen"

2. Fix the header text as the first line of an Excel sheet can be fixed (see file attached)

[fixed header area.xlsx](https://github.com/openkfw/TruBudget/files/5422130/fixed.header.area.xlsx)

header area

pheolink:
Add Fab button at the bottom and scroll back to the top.
New file added ScrollTop.js

Closes #579


Stezido (changes requested):
Good work. Consider the danger.js warnings.

pheolink:
raised a new PR
https://github.com/openkfw/TruBudget/pull/631

PR #632, opened 2020-11-19
Generator test

galethil:
### Checklist

### Description

PR adds generator of Joi definitions from Swagger.
in /api you can use `npm run generate-joi {name_of_the_file_from_root}` e.g. `npm run generate-joi project_update.ts`
This will display the diff between existing definition and generated definition. If you want to rewrite the definition in file just add "r" to the command e.g. `npm run generate-joi project_update.ts r`.

Sometimes there is a need to update Swagger. E.g. if you want to have `.allow('')` in Joi, based on swagger documentation you have to specify `minLength: 0` in Swgger. 
Sometimes you might want custom Joi definition. In that case just add property to swagger e.g. `"x-joi-replace": "MyCustomDefinition"`. Also you can use `x-joi-add` to just add custom definition to the end of generated Joi definition.


api/tsconfig.json
```diff
@@ -11,6 +11,6 @@
     "noUnusedParameters": false,
     "allowSyntheticDefaultImports": true
   },
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "src/scripts/joiGenerator.ts"],
```
Stezido, line 14: isn't "src/**/*.ts" enough to include the "src/scripts/joiGenerator.ts" file too?

Stezido (commented):
Nice structured code, easy to read. The swagger changes and including the swagger into the travis pipeline should be done in separated PRs. Could you create 2 issues:
1. api: Compare/Adjust all joi schemas, swagger schemas and interfaces with the generator script
2. api: Include the validation comparison in a a schema validation job in travis to check if all validation schemas describe the same.

api/tsconfig.json
galethil: you are right, this was added somehow by vscode. Already removed.

PR #638, opened 2020-11-23
e2e: randomizing group name

### LINKED ISSUE #621: e2e-test: Refactor groups.spec
### Description
Currently the groups.spec is failing if the tests are rerun on the same blockchain state again

Rewrite the tests so the tests can be run as often as required using the same blockchain state

### Solution

Randomize group id
Reuse enable/disable user spec function for randonize the userId

### TODO

- [x] Refactor groups e2e-tests so they can be rerun

galethil:
### Checklist

### Description

Randomized group id in group_spec.js so tests can run multiple times.

Closes #621

### DIFF
--- a/e2e-test/cypress/integration/groups_spec.js
+++ b/e2e-test/cypress/integration/groups_spec.js
@@ -1,25 +1,26 @@
 describe("User/Groups Dashboard", function() {
-  const testGroup = { id: "group1" };
-
+  let baseUrl, apiRoute;
   before(() => {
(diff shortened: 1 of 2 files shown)


Stezido (dismissed):
The randomized string is one part of the issue. The other part is to make the tests independeant so if I execute the tests in a different order or execute only some of them, they should work too.

To achieve this every test creates his own group, therefore we need the grioup id randomization

galethil:
@Stezido so the solution I provided is not sufficient? Can you tell in more detail what I should do to complete the task?

daniel-arnauer:
i refactored the test. Should be fine now

georgimld (approved):

PR #642, opened 2020-11-24
e2e: ROOT_SECRET from .env file

### LINKED ISSUE #633: e2e-tests: Add ROOT_SECRET environment variable

## Summary 💡

Currently the passwort of the root user (ROOT_SECRET) is hardcoded and cannot be changed. Adding an env variable to set a custom ROOT_SECRET  would prevent tester issues while setting up the e2e-test locally.

galethil:
### Checklist

### Description

ROOT_SECRET in e2e tests in now configurable using .env file. Just create .env file from  .env.example file. 
You can change the value in the file if you need to.
Variable must start with CYPRESS_ for cypress to be able to read it.

Closes #633


e2e-test/cypress/plugins/index.js
```diff
@@ -115,4 +116,6 @@ module.exports = (on, _config) => {
       return options;
     }
   });
+  _config = dotenvPlugin(_config);
```
Stezido, line 119: Rename the _config  to config since it's used now.

e2e-test/.env.example
```diff
@@ -0,0 +1 @@
+CYPRESS_ROOT_SECRET=root-secret
```
Stezido, line 1: I think we can use the ROOT_SECRET itself. I cannot think of a scenario where the rootsecret of cypress is a different one than the one in the api or elsewhere

Stezido (changes requested):
The documentation has to be updated, but since there is another team meber working on a refactoring of the current contributor documentation you can skip it

e2e-test/cypress/plugins/index.js
galethil: updated.

e2e-test/.env.example
galethil: @Stezido I'm not sure if that is possible since cypress is consuming only env variables starting with `CYPRESS_`

galethil: https://docs.cypress.io/guides/guides/environment-variables.html#Option-3-CYPRESS

Stezido: true, thanks for the information link

Stezido (dismissed):

Stezido (commented):
Please check again if the docker-compose for the e2e-tests is working:
```
sh scripts/testing/start-e2e-tests.sh 
```

Note that the docker-compose e2etest process can be improved by implementing a way to check if it's already provisioned, but it should work anyway

doc/installation/Developer-Setup.md
```diff
@@ -463,11 +465,11 @@ In the `e2e-test` folder you can run the following commands:
 You need to run cypress while also specifying the urls of the frontend, api and excel export service you are using
 
 ```bash
-npm run cypress -- --config baseUrl=http://localhost:3000 --env API_BASE_URL=http://localhost:8080,EXPORT_SERVICE_BASE_URL=http://localhost:8888
+npm run cypress -- --config baseUrl=http://localhost:3000 --env API_BASE_URL=http://localhost:8080,EXPORT_SERVICE_BASE_URL=http://localhost:8888,ROOT_SECRET=root-secret
```
m-janos, line 468: @galethil when you are using .env file, there is no need to use --env flag. The .env file has precedence. There should be mention of two possible runs: One with .env file and the second with --env flag to not cause confusion.

galethil: @Stezido what is the preferred way of running e2e tests? What should be put to Developer documentation?

Stezido: @m-janos yes true but we have to make sure the executor is in the right folder
@galethil If all e2e-test should be executed and only the end result is important the all-in-one-docker-compose should be used. If the docker-compose is already running of course only the required services and the e2e-tests can be restarted. If I only want to test a specific test I would use the ui with npm run cypress or the script with npm run e2e-test:here. We should definitely document the prefered ways for both scenarios:
1. How to run e2e-tests while I can do other stuff and only want the end result.
2. How to run only a set of tests so the try and error cycle is as fast as possible

Stezido: I will only add the all-in-one docker-compose documentation in this PR. The other changes could be done in  https://github.com/openkfw/TruBudget/issues/769

m-janos:
@galethil  After running `sh scripts/testing/start-e2e-tests.sh` Cypress cannot connect to the API.

API container log:
```
Problem with establishing connection to min.io and creating bucket.
Error during searching for bucket Error: getaddrinfo ENOTFOUND nginx
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'nginx'
}
```

Stezido:
@m-janos ty for noticing I added minio to the compose file should work now

m-janos (dismissed):

e2e-test/cypress/integration/subproject_permissions_spec.js
```diff
@@ -6,7 +6,7 @@ const testUser2 = { id: "jxavier", displayname: "Jane Xavier", password: "test"
 const testGroup = { id: "admins", displayname: "Admins" };
 let projectId, subprojectId, permissionsBeforeTesting, baseUrl, apiRoute;
 const subprojectDisplayname = "subproject assign test";
-const rootSecret = "root-secret";
+const rootSecret = Cypress.env("ROOT_SECRET");
 
 describe("Subproject Permissions", function() {
```
m-janos, line 11: One test is failing (locally only) with following error:
```
  1) Subproject Permissions
       Confirmation of multiple grant permission changes grants permissions correctly:
     CypressError: `cy.wait()` timed out waiting `5000ms` for the 3rd request to the route: `listSubprojectPermissions`. No request ever occurred.
```

m-janos (approved):

PR #660, opened 2020-12-02
ui: added groups to batch permission assignment

### LINKED ISSUE #612: ui: Batch editing doesn't include groups for granting permissions
### Description
Currently the batch editing doesn't include groups for granting/revoking permissions, but if using the normal granting/revoking permission dialog it's possible to grant/revoke permissions to groups.
Assigning Groups are not allowed in Trubudgets frontend so that is working correctly in batch editing.

e2e-tests should be added in #404 

### TODO

- [ ] ui: Add groups to all permission user selections in batch editing
- [ ] ui: Be sure that the additional permissions are set correctly for every member in the group (e.g. if granting assign permission to a group view permissions on all resources and for all users in the group has to be set)

galethil:
### Checklist

### Description

Added groups to batch permission assignment. One or more groups could be set for multiple workflowitems.

Closes #612

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Added a close button in user selections [#614](https://github.com/openkfw/TruBudget/issues/614)
 - Added a search field in the validator selection [#639](https://github.com/openkfw/TruBudget/issues/639)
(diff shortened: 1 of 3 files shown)


georgimld (commented):
Please add a new entry to the changelog, rebase the master and then we can merge the PR

georgimld (approved):

PR #668, opened 2020-12-02
ui: added new materialui datepicker

### LINKED ISSUE #629: ui: use new datepicker
since we updated material ui, we can use the material ui datepicker instead of the html native datepicker

keep in mind: the native datepicker returns `undefined ` if not set, the material ui datepicker returns an `empty string`

pheolink:
Added material ui datepicker replacing the old native datepicker.

Close #629


daniel-arnauer (approved):
approved

PR #673, opened 2020-12-07
api:fixing eslint errors

galethil:
### Checklist

### Description

This PR is continuously fixing eslint issues. It can be merged multiple times to prevent huge PR with lot of changes and a lot of conflicts.

No closing of issue until everything is fixed.

### DIFF
--- a/api/.eslintrc.yaml
+++ b/api/.eslintrc.yaml
@@ -57,6 +57,10 @@ rules:
   #   - 2
   #   - always
 
+  "@typescript-eslint/no-unused-expressions":
+  - 2
+  - allowTernary: true
+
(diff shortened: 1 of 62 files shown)


SamuelPull:
@galethil  For improved readability, I suggest deleting the rules from eslintrc which do not need to be overriden, because they are configured in the base by default. E.g. `"@typescript-eslint/lines-between-class-members": 2` has the same effect as not having this rule in the file. I think that encompasses most of our overrides.

`.eslintrc.yaml` is in conflict with master, it might be easiest to copy and paste the master version of this file and than change/delete the rule.

SamuelPull (approved):

PR #749, opened 2021-01-21
ui: add attachment file icon in workflow

### LINKED ISSUE #234: ui: Move validation button for workflow items to workflow item table
Currently, the document validation button is hidden within the workflow item details. Instead, a button (for example paper clip symbol to represent "attachment") could be introduced on the right hand side of the workflow item showing in a more prominent way that a document needs to be validated before closure of the workflow item. This paper clip should only appear if there is a document to validate. Clicking the paper clip should show the document name, who uploaded it and when, and allow to validate the document.

pheolink:
Closes #234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed Actions table of confirmation dialog alignment [#677](https://github.com/openkfw/TruBudget/issues/677)
 
 ### Added
-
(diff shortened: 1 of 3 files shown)


georgimld (changes requested):
You only created tests where it is checked that the attachement icon is NOT displayed, but you should also write a test where the icon should be displayed (maybe in the documents_spec file)

georgimld (approved):

Stezido (dismissed):
Why same tests 2 times? 
Remove it from the create/edit workflowitem specs since it's not matching with create/edit it's more a view thing. 
I suggest to create a new spec named `suproject_view_details.spec`.
Also check e2e-tests they are failing

georgimld (approved):

PR #751, opened 2021-01-21
User selection in group editing

### LINKED ISSUE #725: User selection in group editing has an api call for each selection - SP 0.5
- [x] The issue is present in the latest release.

## Current Behavior 😯

Each time a user is selected or deselected using the checkbox in the dialog, an api call for group.addUser or group.removeUser is bein sent. 

## Expected Behavior 🤔

Users to be added and removed should be stored (e.g. in an array), and the requests should only be sent when clicking the "Done" button.

## Steps to Reproduce 🕹

Steps:

1. open the edit dialog of a group
2. click on the dropdown for selecting users
3. try to select or deselect users and look at what calls are being sent

## Your Environment 🌎

| Tech           | Version |
| -------------- | ------- |
| TruBudget       | v1.17.0  |

SamuelPull:
### Checklist

### Description

This PR adds a new API endpoint for updating of user groups: adding and removing group members in one request (and event.)
This way, it's not necessary to track in frontend redux store which member was added to the group and who was removed. 
After confirming the changes in UI, final membership state is sent to server. Previous state is "overwritten".  Net result is smaller number of network and read-write operations. 

With that, there goes a new `GroupUpdated.Event` and appropriate handlers.

In `api/src/index.ts`, I added a `console.log` to see the error, if any. I have noticed in several other instances, that `logger` used in API doesn't really print out anything meaningful and helpful, no matter the log level.

While making the changes to frontend, I added several eslint rules to enforce the formatting and linting standards. These I tried to formulate in a way, that the rules follow the style already present in the codebase.  As a result, there are some unrelated sections in this PR where formatting/spacing has changed, this is purely for consistency. 

Closes #725


Stezido (changes requested):
Could you please add a description what's this PR about and fix the warnings or at least describe why this warning is happening before I review the PR?

frontend/src/api.js
```diff
@@ -143,7 +153,9 @@ class Api {
       groupId,
       userId
     });
-  listGroup = () => instance.get(`/group.list`);
+  listGroup = () => {
```
SamuelPull, line 156: This can be changed back to arrow func without explicit return, probably there was logging in function body at some point.

frontend/src/pages/Common/UserSelection.js
```diff
@@ -54,6 +54,7 @@ const styles = theme => ({
 });
 
 function UserSelection(props) {
+  const { classes, users, addToSelection, selectedItems, handleDelete } = props;
```
SamuelPull, line 57: I moved props to top of func body, nobody expects to find props deep down in React component.

frontend/src/pages/Users/Dialog.js
```diff
@@ -126,15 +127,15 @@ const Dialog = props => {
...
-            // eslint-disable-next-line no-console
+          // eslint-disable-next-line no-console
           } else console.error("Not a recognized action", action.type);
         });
         hideDashboardDialog();
       };
 
       break;
-    case "editGroup":
-      const group = groups.find(group => group.groupId === editId);
```
SamuelPull, line 137: Renamed `group` to `g`. There is already `group` in outer scope.

SamuelPull:
Rebased onto current master, modified changelog, minor formatting.

Stezido (commented):
@SamuelPull I would expect that the group update can update the whole group so displayname, description, members, permissions and additionaldata. Also in the group update event I should only see what should be updated not the whole group state. so all values mentioned should be optional values like in workflowitem.update.

daniel-arnauer:
These things have to be discussed:

- Why do we need a new intent? Just change the current group.AddUser and group.RemoveUser to group.AddUsers and group.RemoveUsers, so the request does not need a single userId, instead it needs an array of userIds

- Currently, the permission "global.createGroup" is used to be allowed to execute the global.groupUpdate intent. Does this make sense? Is it possible to implement a new permission "global.groupUpdate" because of backwards compatibility?

api/src/service/domain/organization/group_update.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const groupExists = groupExistsResult;
+  if (!groupExists) {
+    return new NotFound(ctx, "group", updateEvent.group.id);
+  }
+
+  // Check authorization (if not root):
+  if (creatingUser.id !== "root") {
+    const intent = "global.createGroup";
+    const globalPermissionsResult = await repository.getGlobalPermissions();
```
daniel-arnauer, line 81: why permission "global.createGroup" ?

daniel-arnauer:
This PR will be closed because this approach is not backwards compatible since here we need a new permission. 
A new, smaller approach with refactoring the current group.addMember and group.removeMember API Calls are done in PR #811

Issue #33, opened 2019-03-22
For changes in a subproject's workflowitem ordering, the history doesn't show _what_ has changed


phpai:
In GitLab by @kevinbader on Aug 14, 2018, 11:25

From #47:

Currently the history item when the order of the workflowitems change looks like: 
```text
mstein changed the workflowitem ordering 
```

It would be nice to get details about what exactly changed.

phpai:
In GitLab by @kevinbader on Aug 14, 2018, 11:25

changed the description

mathiashoeld:
relates to #236

PR #759, opened 2021-01-31
33 workflowitem ordering - history message modification

pheolink:
1. Updated the api to return previous ordering and workflowitems in workflowitem_reordered business event.
2. Updated the frontend history messages for the modified api.

"Closes - #33 "


Stezido:
The implementation is postponed to another release so the PR is closed

Issue #746, opened 2021-01-20
Add connection status to network list - 8SP


Stezido:
## Summary 💡

Add a column to the network table which shows the connection status of a node and if not connected the last seen timestamp

# Subtasks 🔗
 * [ ] Show node connection status -  [PR](https://github.com/openkfw/TruBudget/pull/775)
 * [ ] Display last available timestamp #774

m-janos:
@Stezido Clarification needed. Each network can contain multiple nodes. If connection status for each node should be displayed, the UI has to be modified. Do we have know how it should look like? Some expandable row or just write all node statuses in single row?

m-janos:
Proceed to the subtask after [PR](https://github.com/openkfw/TruBudget/pull/775) is closed

Issue #774, opened 2021-02-11
Add Last Seen Time For Disconnected Node


m-janos:
- [x] I have searched the [issues](https://github.com/openkfw/TruBudget/issues) of this repository and believe that this is not a duplicate.

## Summary 💡
 * When Multichain node is disconnected, display last active timestamp
 * Find the way how to retrieve the time value
 * Screen --> component: e.g. http://localhost:3000/nodes --> Network
 
-------------------------------------------

* [x]  Process runs every 24H on every API
* [x] check if chain connection/permission ready (this applies to not yet connected nodes, on which logging should not even run)
* [x] `getpeerinfo` issued from api returns an array of connected nodes
* [ ] process the array (we do not need all the information)
* [x] Create an event e.g. `store_network_infromation` or such
* [x] Condition: Check if the event for that day is written in the stream
* [x] if not, store the event in stream
* [x] YYYYMMDD as key
* [x] possibly use addresses of nodes as additional keys?
* [x] use also the address of publishing node in keys
* [x] which stream? events in "nodes" are expected to have certain structure and function (get in Nodes.ts) => consult with Stefan (using `network_log` atm)
* [x] all nodes registered in the chain are in the nodes stream. if some of these  nodes are not connected to the chain (`getpeerinfo` returns array with some of them missing), API should take care of this -> look in the stream when was that particular node last seen, and include the information in the reply to frontend

* [x] Frontend gets the information from API - maybe piggyback response in such a way that no breaking change occurs?

## Parent Task 🔗
 * #746

PR #775, opened 2021-02-11
Node Connection Status

m-janos:
### Checklist

### Description
 * Display connection status of each node in the network list (/nodes page)
 * Used `getpeerinfo` to retrieve all connected peers, compare with node list to determine status
 
### Test Plan
 1. in scripts/development/docker-compose.yml add following service right after *master-node*:
 ```
  slave-node:
    build:
      context: ../../blockchain
    environment:
      ORGANIZATION: ${ORGANIZATION}
      RPC_PORT: ${RPC_PORT}
      RPC_PASSWORD: ${RPC_PASSWORD}
      P2P_HOST: master-node
      P2P_PORT: ${P2P_PORT}
      API_PROTO: http
      API_HOST: master-api
      API_PORT: ${API_PORT}
      EXTERNAL_IP: master-node
    volumes:
      - /tmp/bcSlaveVolume:/root/.multichain
    ports:
      - "7448:7447"
 ```
 2. Run app as normal, approve new node and check that Network table on /nodes page **displays** two nodes with **Connection** status 
 3. In new terminal window run `docker exec -it trubudget-dev_slave-node_1 bash`
 4. In afterwards-opened subshell run `multichain-cli TrubudgetChain stop`
 5. Check frontend, one node should be disconnected
 6. Return to the subshell and run `multichaind -txindex -port=7447 -autosubscribe=streams TrubudgetChain@master-node:7447 “” -externalip=master-node`
 7. Refresh the page on the frontend. The node should be connected again

### Resources
 * [Multichain-CLI tutorial](https://github.com/openkfw/TruBudget/blob/master/doc/tutorials/multichain-cli/multichain-cli.md)

### Issue
Subtask of #746

### DIFF
--- a/frontend/src/pages/Nodes/NodesTable.js
+++ b/frontend/src/pages/Nodes/NodesTable.js
@@ -44,7 +44,7 @@ const groupBy = nodes => {
 const NodesTable = ({ nodes, classes }) => {
   const groupedNodes = groupBy(nodes);
   return (
-    <Paper className={classes.paper}>
+    <Paper className={classes.paper} data-test="nodes-table">
       <div className={classes.title}>
         <Typography variant="h6" color="primary" id="tableTitle">
           {strings.nodesDashboard.network}
--- a/frontend/src/pages/Nodes/NodesTable.js
+++ b/frontend/src/pages/Nodes/NodesTable.js
@@ -29,11 +29,12 @@ const groupBy = nodes => {
     if (node.currentAccess.accessType !== "none") {
       const key = node.address.organization;
       const index = acc.findIndex(x => x.organization === key);
+      const permissionsString = `${node.currentAccess.accessType} (${node.isConnected ? strings.status.connection : strings.status.not_connected})`;
       if (index === -1) {
-        acc.push({ organization: key, count: 1, permissions: node.currentAccess.accessType });
+        acc.push({ organization: key, count: 1, permissions: permissionsString});
       } else {
         acc[index].count += 1;
-        acc[index].permissions = acc[index].permissions.concat(`, ${node.currentAccess.accessType}`);
+        acc[index].permissions = acc[index].permissions.concat(`, ${permissionsString}`);
       }
     }
     return acc;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Fixed -->
 
+## [unreleased]
+
+### Added
+
+- Display connection status for each Multichain node [#746](https://github.com/openkfw/TruBudget/issues/746)
+
 ## [1.19.0] - 2021-02-10
 
 ### Added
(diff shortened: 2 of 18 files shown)


frontend/src/pages/Nodes/NodesTable.js
```diff
@@ -29,11 +29,12 @@ const groupBy = nodes => {
     if (node.currentAccess.accessType !== "none") {
       const key = node.address.organization;
       const index = acc.findIndex(x => x.organization === key);
+      const permissionsString = `${node.currentAccess.accessType} (${node.isConnected ? strings.status.connection : strings.status.not_connected})`;
```
georgimld, line 32: Might be a better idea to show the connection status not as a string but as a green/red dot, like in the status page (`frontend/src/pages/Status/StatusTable`)

georgimld:
Found a possible bug: 
When a slave node is trying to connect to the network, it is retrying every 10 seconds and the logs in the container are

"Please ask blockchain admin or user having activate permission to let you connect and/or transact:
multichain-cli TrubudgetChain grant `addressOfNode` connect
multichain-cli TrubudgetChain grant `addressOfNode` connect,send,receive"

When approving a new node (Step 2 in the test plan), the node is given permissions (check using the multichain-cli command `listpermissions * addressOfNode`), but the node still appears offline (`getpeerinfo` shows no node) and according to the logs in the node's container the slave node keeps restarting every 10 seconds - which is why it appears not connected, even though it does have the necessary permissions to connect.

frontend/src/pages/Nodes/NodesTable.js
m-janos: @georgimld, @Stezido text representation is simple and do the thing, of course the table would look more perky with circles but displaying multiple circles on the same row would look "dirty" with greater count of nodes. I propose to put each node on separated node and create new column Connection to look like in Status table then. What do you think?

georgimld: @m-janos I also think it would be a good idea to display each node on a separate row. We already discussed this in yesterday's call with @Stezido and @SamuelPull and it was suggested for the "last seen" issue, right?

georgimld: maybe like this + the "last seen" timestamp in another column eventually?

Screenshot 2021-02-18 at 14 42 41

m-janos: @georgimld all right, this will be resolved in #774 and #785

m-janos:
> Found a possible bug:
> When a slave node is trying to connect to the network, it is retrying every 10 seconds and the logs in the container are
> 
> "Please ask blockchain admin or user having activate permission to let you connect and/or transact:
> multichain-cli TrubudgetChain grant `addressOfNode` connect
> multichain-cli TrubudgetChain grant `addressOfNode` connect,send,receive"
> 
> When approving a new node (Step 2 in the test plan), the node is given permissions (check using the multichain-cli command `listpermissions * addressOfNode`), but the node still appears offline (`getpeerinfo` shows no node) and according to the logs in the node's container the slave node keeps restarting every 10 seconds - which is why it appears not connected, even though it does have the necessary permissions to connect.

A new node is registered as connected after some time (one case takes 14 Minutes) and the "Please ask blockchain admin" message will disappear by itself.

georgimld (approved):

Issue #659, opened 2020-12-01
Github actions improvements 


Stezido:
- [x] I have searched the [issues](https://github.com/openkfw/TruBudget/issues) of this repository and believe that this is not a duplicate.

## Summary 💡

Currently github actions work but following improvements can be made

### TODO

 - [ ] build/push docker images with github action instead via script
 - [ ] only run necessary jobs (e.g. if I changed the frontend only - only frontend should be audited/tested for the pull_request)
 - [ ] rename env vars on e2e-test ci-cd side 
 - [ ] add linting jobs

## Motivation 🔦

The Todos provide consistency/ faster pipeline execution

Stezido:
Following subtask should be completed to ensure linting jobs are working properly:
 -  Fix eslint issues #787

daniel-arnauer:
Closed in #794

PR #787, opened 2021-02-23
Fix all eslint issues in all the projects

pheolink:
Sub task of "#659"

Fixes all eslint issues, so lint-ci  github job will work correctly.


daniel-arnauer (approved):
approved

Issue #799, opened 2021-03-11
Document storage: Sign saved file hash


Stezido:
- [x] I have searched the [issues](https://github.com/openkfw/TruBudget/issues) of this repository and believe that this is not a duplicate.

## Summary 💡
When a file is uploaded only the docIds are saved onto the blockchain.
In terms of security this is not enough because the file located at the accesspoint can still be modified. 
To be sure that the downloaded file behind the URL is still the same a digital signature must be uploaded too. This signature can then be checked by the api if the file downloaded is the expected one.

PR #809, opened 2021-04-12
document feature implementation

Stezido:
### Solution?
document_feature

0. publish public key
     - if document service enabled
     - pass required env vars "PUBLIC_KEY_PATH", "PRIVATE_KEY_PATH" - must be the same for nodes from same orga
     - publish public key to public_keys-stream
1. upload
    - api trubudget
        - createWorklfoiwtem (endpoint)
            - documentUpload (api service)
                1. check and generate a unique docId
                2. call upload endpoint of storage service (docId)
                2. "offchain_documents" stream - create document_upload event (docId, filename, orga)
                3. Check orga access -> check orga public keys (own public key included) -> encrypt secrets with key and generate event per orga on document_secrets stream (docId, orga, encrypted secret)
                4. Generate event on document_access_point stream to share where document can be downloaded (orga, accesspoint)
        - updateWorkflowitem
            - documentUpload (same as above)
    - storage-service
        - upload (return objectId, generated secret)
           1. generate secret
           2. store object on connected storage
           3. return success with objectId and secret
2. list
    check eventsourcing for offchain-documents
    check workflowitem.list, subproject.viewDetails
3. download
    - api trubudget
        - downloadDocument (endpoint)
            1. getSecret(): Check document of offchain_documents stream - check docId - get secret of my orga with docId 
            2. getAccessPoint(): Check document of offchain_documents stream - check orga- get access point of this orga
            2. Decrypt secret with own private key
            1. call download endpoint of storage service using access-point/[docId] (passing decrypted secret)

### Open for discussion

- [ ] workflowitem create - should workflowitem still be created after document upload fail? - Rollback topic in TODOs

### TODOs

- [x] publish public key including event sourcing
- [x] storage service setup including minio as storage (upload(objectId)->(docId, secret), download(docId in query, secret)->(object), secret generation on upload)
- [x] Trubudget API - generate document upload api service (2 events - document_uploaded, secret_published (for own orga))
- [x] Trubudget API - generate document share api service (event - secret_published for 1 organization)
- [x] Trubudget API - provide connection/client to/for storage service internal(connToken) and external(env var)
- [x] Trubudget API - adapt workflowitem create with new doc upload service
- [x] Trubudget API - adapt workflowitem update with new doc upload service
- [x] Trubudget API - adapt workflowitem grant permissions with new doc share service
- [x] document download via accesspoint
- [x] upload document -> calculate and store hash on chain 
- [x] check organization_url on api startup if changed publish an organization_url_updated event (better naming needed)
- [x] refactor document event sourcing (template groups eventsourcing) - parallel to organization_url_updated 
- [x] use private key of organization to ensure public key exists / generate and push public key
- [x] update getVersion endpoint (add storage service to status page) 
- [x] adapt nginx for storage service enabled trubudget (configureServer.sh)
- [x] adapt frontend document info table including e2e-tests
- [x] check backwardscompatability and test it if possible
- [x] test an fix file upload without storage service with a bigger file 
- [x] create deployment with 2 organizations/ test the doc feature with 2 orgas @georgimld @Stezido 

### TODOs after beta version release
- [ ] write documentation
- [ ] Check which e2e-tests would make sense
- [ ] finalize unit tests
- [ ]  Rollback of document upload in case of error
- [ ] publish a digital signature with the document on chain so it can't be modified (check it on download)#799 

#### Events in offchain_documents
Events:
- document_uploaded
```
{
  docId,
  fileName,
  organization, (doc publisher)
}
```
- secret_published
```
{
  docId,
  organization, (encrypted for)
  encrypted secret
}
```
- storage_service_url_published
```
{
  organization, (doc publisher)
  url
}
```

Bug Table

|  short Bug name| Bug description          | critical for release|
| ------------------ |------------------------| -------------------|
|                            |                                    |                           |


Stezido:
since we are using eventsourcing in streams we don't need to create that many streams. We just need the public_keys stream and the document stream (offchain_documents) where all document related events can be saved in offchain_documents stream.

Stezido:
we should impement an env var to enable the storage service and the doc feature, so we can ensure that this change isn't a breaking one. If the doc service is enabled private and public key must be provided. 
We also shouldn't remove the old way documents are stored because of backwards compatiblity

Stezido:
The API needs to know the external URL of the storage service so it can be stored on blockchain. (organization URL) 
This will be the frontend url - the frontend (nginx) handles the networking to the storage service.

Stezido:
What happens if a frontend of a node or a node itself shuts down? Are the documents unavailable then?
Do we provide a possiblity to scale the storage service so multiple nodes of an organization can support a storage service connected to the **same** storage. in that case an array of org urls must be stored on blockchain.

Stezido:
Is it a use case to use the document feature (download docs) while not having an own storage service?
If yes we need to be able to enable docs feature and a seperate enable env var to enable storage service

Stezido:
divide download of documents in storage service in external and internal and only map the external endpoint in the frontend's nginx

Stezido:
There's a keytest folder in api to provide test public/private key to the api to start the api with document feature enabled execute folowing commands:
```
cd api
export ORGANIZATION=KfW
export ROOT_SECRET=root-secret
export ORGANIZATION_VAULT_SECRET=secret
export DOCUMENT_FEATURE_ENABLED=true
export ORGANIZATION_PUBLIC_KEY=$(cat ./keytest/rsa.public | base64)
export ORGANIZATION_PRIVATE_KEY=$(cat ./keytest/rsa.private | base64)
export STORAGE_SERVICE_EXTERNAL_HOST=localhost
npm start
```

If you want to use the document feature a storage service is needed and has to be started:
defaults: 
 - host: localhost
 - port: 8090
```
cd storage-service
npm start
```
to specify host/port of storage service for the api use following commands before starting the api:
```
export STORAGE_SERVICE_HOST=localhost
export STORAGE_SERVICE_HOST=8090
```

Stezido:
Can we use the organizations walletaddress/ vault / private key which is published via the orga's stream to generate the public key?
That way an orga must not provide an organization_vault_secret, a private key and a public key. It would be suffient to provide the organization_vault_secret which is used to encrypt the orga's stream where a generated private key is found. When saving the private key the only thing the api has to add is to ensure the public key is also published.

georgimld:
To start the storage service you need to specify some environment variables. You can do this in two ways

- by starting the whole app with docker
```
cd storage-service
docker-compose up
```

- or locally with:
```
cd storage-service
export STORAGE_HOST=storage
export STORAGE_PORT=8090
export ACCESS_CONTROL_ALLOW_ORIGIN="*"
export MINIO_ACCESS_KEY= minio
export MINIO_SECRET_KEY= minio123
export MINIO_PORT=9000
export MINIO_HOST=minio
npm start
```

georgimld:
Latest architecture decisions:

- The url of the storage system (for the download) will be saved on the `offchain_documents` stream in the `storage_service_url_published`event
- On api startup, the url is saved (if it doesn't exist yet) or updated (if it is different from the one stored on the blockchain)
- we implemented an event sourcing (`storage_service_url_eventsourcing.ts`) that saves the current url in the document infos when performing the document eventsourcing (`document_eventsourcing.ts`)

Stezido:
multichain generates an own public key per wallet address/organization which can be obtained by `getaddresses verbose=true` (multichain-cli)
That way we don't need the public_keys stream and the whole public key publish/update stuff

georgimld (approved):

PR #820, opened 2021-05-05
api: change user id to 64 byte

### LINKED ISSUE #266: Increase the size of User.ID field (currently 32-char) to 64-char
During our tests, we noticed that we can not fill the User ID with one of our client's email (monique.bendahan@fas-amazonas.org). 

For this reason, we kindly ask to increase the size of the field to support 64 characters.

Thanks in advance.

daniel-arnauer:
Issue #266


daniel-arnauer:
Closed because now there is no need to change this

PR #880, opened 2021-07-06
provisioning: fix filenames

daniel-arnauer:
The provisioning was faulty due to changes on the document upload in the API. 

- [x] Change id to fileName
- [x] Allow double using of the same filenames

### DIFF
--- a/frontend/src/pages/Documents/DocumentUpload.js
+++ b/frontend/src/pages/Documents/DocumentUpload.js
@@ -29,10 +29,7 @@ export default class DocumentUpload extends Component {
           <DocumentOverview documents={workflowDocuments} validationActive={false} />
         </div>
         <div style={{ display: "flex", flexDirection: "row", justifyContent: "center" }}>
-          <Button
-            style={styles.uploadButton}
-            component="div"
-          >
+          <Button style={styles.uploadButton} component="div">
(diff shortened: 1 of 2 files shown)


provisioning/src/index.js
```diff
@@ -443,36 +437,12 @@ async function testWorkflowitemUpdate(folder) {
...
-        },
-      ],
-    });
-    throw Error(
-      "Updated an existing document, but that isn't supposed to work :("
-    );
-  } catch (_err) {
-    // ignore error
-  }
-
```
daniel-arnauer, line 475: this is obsolete. Moreover, these business rules should not be here, it should be an API test

frontend/src/pages/Documents/DocumentUpload.js
```diff
@@ -48,7 +45,9 @@ export default class DocumentUpload extends Component {
                       storeWorkflowDocument(dataUrl, file.name);
                     }
                   };
-                  reader.readAsDataURL(file);
+                  if (file) {
+                    reader.readAsDataURL(file);
+                  }
```
daniel-arnauer, line 50: if cancel is pressed in the file explorer, `file` is `undefined`

Stezido (approved):

PR #906, opened 2021-08-04
ui: hotifx

daniel-arnauer:
Fix e2e-test status_spec.js in master


frontend/.gitignore
```diff
@@ -25,6 +25,3 @@ cypress/videos
 # certificates folder
 certificates/*
 !certificates/Readme.md
-
-# dotenv environment variables file
-.env
```
daniel-arnauer, line 30: .env file is needed to obtain the value of $npm_package_version.

other solutions:  

- get npm_package_version directly from package.json => security concern since the client will receive the opackage.json file
- use a node module
- Set an env-variable  that obtains $npm_package_version everytime the start or build script start (looks hacky)

Stezido (changes requested):
I tried it out , but it doesn't work - frontend service is showing `?` instead of the version.
I started the env via the development start script:
```
sh scripts/development/start-dev.sh --slim --build
```

georgimld (approved):

PR #917, opened 2021-08-06
Improve logging in the api

### LINKED ISSUE #916: Improve logging in the api

## Description 😯

The api needs to have a better error logging in order to provide more info when something goes wrong.

### Example: 
For example in this case no explanation was given as to why the api couldn't create the organization stream. After much debugging it was found that the ORGANIZATION_VAULT_SECRET was not correct - this should have been logged appropriately
Screenshot 2021-08-06 at 13 50 47

georgimld:
### TODOs

- [ ] log the error in case the **organization vault secret** is incorrect ( to reproduce: start api with an organization vault secret, then stop the api and restart it with another one)

Closes #916


mayrmartin:
Will be addressed via #940

PR #949, opened 2021-08-25
ui: fix reject-icon for workflowitems

daniel-arnauer:
Fix the icon in the workflowitem-details dialog. Also replace the icon with a better one.


laurenzhonauer (approved):
Looks Good!

PR #953, opened 2021-08-26
Fix minio and storage-service synchronization

### LINKED ISSUE #946: storage-service: Error "The specified bucket does not exist"

## Description 😯

If minio restarts and is not persisting the data/bucket the storage-service doesn't create a new bucket before putting object onto the storage which lead into an error.

### How to reproduce 🕹

Start trubudget with storage service/ document feature enabled and restart minio without persiting data.
Then try to upload a document while creating a workflowitem.

### Expected behaviour

The storage service should create a new bucket when it gets the error not only on startup.
I would not checking it before each call due to performance but I think reacting with new bucket creation including a log info is fine.

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| Frontend        | v1.24.0 |
| API             | v1.24.0  |
| Blockchain      | v1.24.0 |
| Storage-Service | v1.24.0  |

daniel-arnauer:
### Description

When restarting minio without persisted data, the storage-service does not re-create the bucket (its like a database-name) on minio, which leads to an error. 

### Solution

Storage-service should always check if the bucket exists on minio. If it does not exist, re-create it.

### Reproduce the problem
To reproduce the problem (for testing), see #946

Closes #946

### DIFF
--- a/storage-service/src/minio.ts
+++ b/storage-service/src/minio.ts
@@ -57,7 +57,7 @@ const makeBucket = (bucket: string, cb: Function) => {
   });
 };
 
-export const makeBucketAsPromised = (bucket: string) => {
+const makeBucketAsPromised = (bucket: string) => {
   return new Promise((resolve, reject) => {
     makeBucket(bucket, (err) => {
       if (err) return reject(err);
@@ -74,7 +74,7 @@ const upload = (
   cb: Function,
 ) => {
   const s = new Readable();
-  s._read = () => {};
+  s._read = () => { };
   s.push(content);
   s.push(null);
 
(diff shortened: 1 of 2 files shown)


storage-service/src/minio.ts
```diff
@@ -199,16 +201,23 @@ const establishConnection = async () => {
...
+      }
+      if (buckets.length === 0) {
+        await establishConnection();
+      }
+      resolve(true);
+    });
   });
 };
 
-if (config.storage.host) {
```
daniel-arnauer, line 208: This condition  is useless, since config.storage.host has a default value

frontend/src/sagas.js
```diff
@@ -1719,7 +1789,20 @@ export function* fetchNextWorkflowitemHistoryPageSaga({
...
+        type: SNACKBAR_MESSAGE,
+        message: error.message
+      });
+      yield put({
+        type: SHOW_SNACKBAR,
+        show: true,
+        isError: true,
+        isWarning: false
+      });
+    }
```
daniel-arnauer, line 1805: This is a hacky way to show the red error snackbar if the download fails. Normalle, `handleError()` should be called but this requires refactoring of the storage-service error handling

Stezido (approved):

PR #976, opened 2021-09-08
Add check for user and group name

### LINKED ISSUE #943: Assign permission dialog assigns permission to user and group

## Description 😯
Workflowitem assign permission assign permission to user and group when a user & a group have the same name.  The API does not assign additional permissions (case workflowitem)

1. Create a user
2. Create a group with the same name as the username (& assign the user to the group)
3. Choose any project / sub-project / workflowitem
4. Assign permissions to the user or the group. Both fields for user and group will be checked in the permission dialog

georgimld:
### Checklist

### Description
It should not be possible to create a group or a user with an id that is already taken (by a user or a group). So far the check was only done for users on user creation and for groups on group creation. Now both are checked in either case

Closes #943

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Removed -->
 
-<!-- ### Fixed -->
+### Fixed
+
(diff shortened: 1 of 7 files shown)


daniel-arnauer:
This PR does not resolve the bug from the permission confirmation dialog that is mentioned in #943 if a group-id and user-id already exist with the same value. This should be resolved as well.

georgimld:
> future

A way to solve this for situations where a user and a group with the same id already exist in a network is to manually disable the user

daniel-arnauer (approved):

PR #992, opened 2021-09-29
Fix swagger link

daniel-arnauer:
### Checklist

- [x] Fix swagger URL http://localhost:8080/api/documentation/static/index.html
- [x] Add e2e-test

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -16,7 +16,7 @@
         "fastify-helmet": "^5.3.2",
         "fastify-jwt": "^2.3.0",
         "fastify-metrics": "^6.0.3",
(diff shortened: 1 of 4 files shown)


laurenzhonauer (changes requested):
The guys that made fastify-helmet (which we use for content-security-policy) and fastify-swagger also stumbled across our issue and proposed a solution that seems a little bit safer than ours.

Please have a look at the [fastify-swagger Readme](https://github.com/fastify/fastify-swagger/blob/master/README.md#integration) and try to use a similar config.

laurenzhonauer (approved):

PR #997, opened 2021-10-04
api: fix npm vulnerabilities

laurenzhonauer:


email-notification-service/src/db.ts
```diff
@@ -126,10 +126,10 @@ class DbConnector {
     return "";
   };
 
-  private initializeConnection = (): knex => {
+  private initializeConnection = (): Knex => {
     logger.info("Initialize database connection");
     logger.info(config);
-    const knexConfig: knex.Config = {
+    const knexConfig = {
```
Stezido, line 132: is there no config type anymore?

laurenzhonauer: Yes there is.

I had to experiment with the types a bit, this one got lost in the process. I pushed again with Knex.Config Type

Stezido (approved):

PR #1001, opened 2021-10-18
scripts: update dev and operation setup

### LINKED ISSUE #979: Update docker developer setup 3 SP
Update docker developer setup for more features and easier developing
This is a follow up of #904 and #905
## Description 💡

- [x] Add option `--enable-service [services...] `
- [x] Only use one `.env_example` file and set `env` vars the same ways as in the operation setup
- [ ] ~~Add option `--production`. This will enable `nginx` (the same way as in operation setup but with local builds instead of pulling images from dockerhub)~~
- [x] When using --build, only the selected services are rebuild (now, all services in docker-compose are rebuilt)
- [x] Add `--down` to stop and remove docker containers

daniel-arnauer:
### Description
Refactored the dev bash script for more functions and easier handling of env vars

Closes #979


mayrmartin (approved):

mayrmartin:
@laurenzhonauer pls check if the change from sed to perl works on windows machines with wsl2 (without additional installing perl)

PR #1018, opened 2021-11-23
api is not exiting on unknown Event type

### LINKED ISSUE #1017: Not implemented Event type on chain breaks api

## Description 😯

If Trubudget's api reads an event type which is not implemented it exits with error code 1. I agree that this should be an error/warning, but it should not lead to an program exit.

### How to reproduce 🕹

Store an blockchain event with an unknown event type and start the api

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| API             | v1.27.0  |
| Blockchain      | v1.27.0  |

Stezido:
### Checklist

### Description
This PR is only a workarround. The main issue is that documents stored via mutlichains offchain mechanic are stored in trubudget's cache. In a future fix these documents should not be part of the cache anymore.
Tasks done in this PR:
- Skip and log errors happening while converting to readable items
- Adapt error messages including all necessary information
- Eventsourcing of workflowitems skip the workflowitem_document_uploaded event since document uploads are handled via workflowitem_update event

### Additional Change

It was not possible to exclude certain vulnerabilities to pass the audit checks if there is no fix for the vulnerability yet.
With the new package better-npm-audit you can exclude vulnerabilities with a command argument or with a file called `.nsprc`.

Closes #1017

### DIFF
--- a/.github/workflows/audit/action.yml
+++ b/.github/workflows/audit/action.yml
@@ -3,5 +3,5 @@ runs:
   using: "composite"
   steps:
     - name: Audit project ${{ matrix.project }}
-      run: cd ${{ matrix.project }} && npm install --no-audit && npm audit --production
+      run: cd ${{ matrix.project }} && npm install --no-audit && npm run audit -- --production
       shell: bash
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "axios": "^0.21.4",
         "bcryptjs": "^2.4.3",
+        "better-npm-audit": "^3.4.0",
         "fastify": "^3.11.0",
         "fastify-cors": "^6.0.2",
         "fastify-helmet": "^5.3.2",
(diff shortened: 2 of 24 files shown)


.github/workflows/github-ci.yml
```diff
@@ -45,17 +45,16 @@ jobs:
     strategy:
       matrix:
         node-version: [10.x]
-        project:
-          [
+        project: [
             frontend,
             api,
-            blockchain,
+            # blockchain,
```
mayrmartin, line 51: Consider removing if not used

frontend/docker-compose.yml
```diff
@@ -1,37 +1,43 @@
 version: "3.2"
 services:
   master:
-    image: trubudget/blockchain:latest
+    image: trubudget/blockchain:v1.27.0
```
mayrmartin, line 4: Why lock the TB version to 1.27? Also, why are the ports set to a fixed value & why is provisioning disabled in the docker-compose.yml?

Stezido: sorry didn't want to push this file is was for testing purposes

.github/workflows/github-ci.yml
Stezido: currently the blockchain project has some moderate vulnerabilities so the pipeline wouldn't go through.

georgimld (approved):

PR #1044, opened 2022-02-03
Burkina faso project features

BorisFromBE:
### Checklist

- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/openkfw/TruBudget/blob/master/CONTRIBUTING.md).

### Description
As part of the integration of Trubudget in Burkina Faso and its integration with the local Information System, some functional improvements were identified and developped. One of those improvement consisted in adding in a structured way a new field "responsibleOrganization" at project level that was missing in the global data model. This corresponds to the local entity (ministry, administration,etc.) in charge of the project. 

In this request you will find:
-Modifications to the Project class to comply with Burkina Faso requirements(having a **responsible organisation** for each project)
-Modifications to api requests interacting with projects
-Modifications to the frontend so that a responsible organisation can be specified when creating a project - We added this so you can interact with our modifications via the frontend. This feature is NOT part of our SIGFE application in production because creation is not handled via the frontend in our case. 
--Modifications to the frontend so the responsible organisation is displayed when opening the Subproject page. 

We believe this notion of responsible organization makes sense in a variety of contexts and is therefore worth adding to the source code of TruBudget. This will also help reduce the gap between Trubudget and SIGFE.

Closes ModifyMe


Stezido:
Thank you for contributing! Does the additionalData field on project level fit your requirements? Since you are not using the frontend of Trubudget it's only important that you can pass the extending the `responsibleOrganization` field to each project. The additionalData field was built for such use-cases.

PR #1096, opened 2022-04-07
reject backups before 2.0

### LINKED ISSUE #1072: bc: Delete md5 hashing from restore
### Discussed in https://github.com/openkfw/TruBudget/discussions/870

<div type='discussions-op-text'>

<sup>Originally posted by **georgimld** December 16, 2020</sup>

## Summary 💡

When creating a backup, a hashing is now calculated with a sha256 hash. It used to be created with md5, but as mentioned in #513  it wasn't working perfectly. Restoring a backup is now possible with both the old version (md5 hash) and the new (sha256), as implemented in #693 . 
Deleting the md5 hash function would not allow old backups to be restored anymore, this would be a breaking change and should only be implemented in Trubudget 2.0.0.
</div>

### LINKED ISSUE #1076: Modify backup endpoint to not accept versions older than TB 2.0
## Description 💡

Starting TruBudget 2.0, older backups will not be allowed anymore, so the `./backup`endpoint in the `blockchain` service must be updated accordingly

mayrmartin:
### Description
- Add: Reject backups that have not been created with the same major release version as the current blockchain deployment
- Remove: md5 hashing

Closes #1076 
Closes #1072


blockchain/check-backup.js
```diff
@@ -67,23 +66,35 @@ stream.on("finish", async () => {
...
+      config.Version.split(".")[0] === version.split(".")[0];
+
     if (isValidMetadataFile) {
-      console.log(`The provided backup file is valid\n`);
-      console.log(`No updated backup is created`);
+      console.log("The provided backup file is valid\n");
+      console.log("No updated backup is created");
       process.exit(0);
+    } else if (compatibleVersions) {
+      console.log("The provided backup is from a prior major version.\n");
```
daniel-arnauer, line 80: So the Versions are compatible according to line 79 but they are not in line 80?

blockchain/check-backup.js
```diff
@@ -67,23 +66,35 @@ stream.on("finish", async () => {
...
+    //Check for major version compatibility
+    const compatibleVersions =
+      config.hasOwnProperty("Version") &&
+      config.Version.split(".")[0] === version.split(".")[0];
+
     if (isValidMetadataFile) {
-      console.log(`The provided backup file is valid\n`);
-      console.log(`No updated backup is created`);
+      console.log("The provided backup file is valid\n");
+      console.log("No updated backup is created");
```
daniel-arnauer, line 77: Try not to use else statements when possible: Instead, check with if and do a process.exit(1)

e2e-test/cypress/integration/backup_spec.js
```diff
@@ -108,7 +108,7 @@ describe("Backup Feature", function() {
             expect(success).to.eq(true);
           });
           cy.get("[data-test=client-snackbar]")
-            .contains("Not a valid TruBudget backup")
+            .contains("failed to restore backup: Backup with these configurations is not permitted")
             .should("be.visible");
           cy.url()
             .should("include", "/projects")
```
daniel-arnauer, line 114: Add a some tests for the new feature

daniel-arnauer (approved):

PR #1206, opened 2022-06-22
1105 fix confirmation error

### LINKED ISSUE #1105: Frozen pop-up after permission

## Description 😯
When giving permissions to a group (probably on workflowitem level), the pop-up remains frozen after clicking "Confirm", the only way to exit the pop-up is to click on "Cancel", as seen in the screenshot below:

image

However, after clicking cancel and checking, it was noticed that the permission granting actually worked. 
So the problem probably is that the "Confirm" button doesn't trigger a confirmation message and doesn't close the pop-up.

### How to reproduce 🕹

To reproduce the bug you need to force the frontend to return an error, for example by going to `frontend/src/sagas.js` in the function `assignProjectSaga` and adding `throw new Error();` in the `try`block  - don't forget to remove this once the fix is done!

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| Frontend        | v1.28.1  |
| API             | master  |
| Blockchain      | v1.28.1   |
| Export-Service  | v1.28.1   |
| Email-Service   | v1.28.1   |
| Storage-Service | v1.28.1  |

daniel-arnauer:
### Description

Display an error message in the confirmation dialog if an API request returns an error.

- [x] Save the error message in redux store
- [x] Display the error message with an error icon in the confirmation dialog. The Error message should be something like: "Error: [error_message_grom_API]. Please try again later"
- [x] The Action counter (bottom left) should not be visible if there is an error
- [x] The Buttons (bottom right) should not be visible: Instead show another Button with the label "Close"

Closes #1105


frontend/src/pages/Confirmation/DialogButtons.js
```diff
@@ -68,6 +70,18 @@ const DialogButtons = props => {
...
+
+      {hasFailure && (
+        <Button
+          disabled={executingActions || actionsAreExecuted}
+          aria-label="cancel"
+          data-test="confirmation-dialog-cancel"
+          color={submitable ? "secondary" : "primary"}
+          onClick={() => onCancel()}
+        >
+          Close
```
georgimld, line 82: you should use a translation here for "Close" -> strings.common.close

georgimld (changes requested):
The changes you did so far are good. However, the bug is currently fixed only on the project level and it would be great if it would be fixed on all levels (so also subproject and workflowitem). 
So TODO
- [ ] same behaviour for an error in assigning subprojects (tipp: action ASSIGN_SUBPROJECT_FAILURE)
- [ ] for error in assigning workflowitems (ASSIGN_WORKFLOWITEM_FAILURE)

frontend/src/pages/Confirmation/reducer.js
```diff
@@ -291,7 +292,8 @@ export default function confirmationReducer(state = defaultState, action) {
     case DISABLE_USER_FAILURE:
     case CREATE_WORKFLOW_FAILURE:
     case CREATE_SUBPROJECT_FAILURE:
-      return defaultState.set("open", true);
+      console.log(action);
```
georgimld, line 295: this console.log should be deleted

georgimld:
> The changes you did so far are good. However, the bug is currently fixed only on the project level and it would be great if it would be fixed on all levels (so also subproject and workflowitem). So TODO:
> 
> * [x]  same behaviour for an error in assigning subprojects (tipp: action ASSIGN_SUBPROJECT_FAILURE)
> * [x]  for error in assigning workflowitems (ASSIGN_WORKFLOWITEM_FAILURE)

this is already fixed, so nevermind

georgimld (approved):

PR #1208, opened 2022-06-23
use unicode regexp for validation schemas

### LINKED ISSUE #1202: Patterns are too restrictive for names of projects/workflowitems SP1

## Description 😯
the pattern for displayName in project and workflowitem (and probably also subproject) is too restrictive. 
In particular, some orthographic characters are missing, such as:

-  the en dash (–), 
- orthographic quotation marks (»«„“”''''), but also 
- the space without a break (&nbsp; in HTML). 

Possibly a Unicode RegExp could help here, e.g.: /^([\p{L}\p{N}\p{P}\p{M}\p{S}\p{Zs}]*)$/ and This would also be helpful to be able to use TruBudget with other writing systems (Arabic, Hebrew, Amharic, Farsi, ...) that are covered by these Unicode classes.

### How to reproduce 🕹
Try to create a project or workflowitem with the above mentioned characters

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| TruBudget        | v2.0.0  |

Stezido:
### Description

Change ascii regexp to Unicode RegExp using xregexp since joi doesn't support unicode properties like pL/p{Letter}

Closes #1202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ### Added -->
 
-<!-- ### Changed -->
+### Changed
+
+- Use unicode properties for regex validation schemas [#1202](https://github.com/openkfw/TruBudget/issues/1202)
 
 <!-- ### Deprecated -->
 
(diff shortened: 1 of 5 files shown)


Stezido:
e2e-test are passing. The provisioning was lasting longer than the provisioning check timeout so we have to increase it.

PR #1234, opened 2022-07-20
Blockchain wallet import

### LINKED ISSUE #1061: Migration 1.0 to 2.0
### Discussed in https://github.com/openkfw/TruBudget/discussions/1023

<div type='discussions-op-text'>

<sup>Originally posted by **mayrmartin** December  6, 2021</sup>
We are planing for TruBudget 2.0. In this context, we aim to solve a GDPR related issue when storing documents via on-chain storage. At the moment - depending on the system configuration - data is either stored on the off-chain or on the on-chain storage. When storing on-chain documents can not be deleted in the future and documents get shared with all nodes in the network. To address this problem, we will **discontinue** the on-chain storage option. This means existing files will not be further stored on the chain but rather on the storage service.

This - obviously - will cause some breaking changes. To allow a smooth transition from TruBudget 1. x to 2. x, we will provide a migration script. When developing this migration script, some considerations to take into account:

- Where to read/write the data from/to? API/Blockchain?
  - We will read data from the Blockchain and use the API to write the data on the new chain
- Documents
  - Documents stored off-chain should be stored on storage service
  - Documents stored on-chain should be stored on storage service
  - What happens with secrets related to documents?
- User
  - Can be migrated 1:1
- Chain encryption
  - Data on a new chain should be encrypted if the old chain was encrypted
- Timestamps
  - should migrated events have the original timestamp? or shouldn't there be an additional timestamp for not migrated events?

</div>

mayrmartin:
### Description
Adds import wallet functionality to the blockchain API in order to import wallets from 1.x into 2.x

Closes #1061


Stezido (approved):

sgtroy88:
https://github.com/openkfw/TruBudget/commit/eb0d5f6a85a7938aa30767a953cca417f58248bc

Issue #1089, opened 2022-04-06
Check versions of dependencies/components SP 1


Stezido:
- [ ] I have searched the [issues](https://github.com/openkfw/TruBudget/issues) of this repository and believe that this is not a duplicate.

## Description 😯

For trubudget 2.0 every possible major update should be done in every component of Trubudget

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| Frontend        | v1.?.?  |
| API             | v1.?.?  |
| Blockchain      | v1.?.?  |
| Export-Service  | v1.?.?  |
| Email-Service   | v1.?.?  |
| Storage-Service | v1.?.?  |

Stezido:
TODOs:
Check dependabot and implement an automated solution to check for dependency deprecation or updates.
Implementation of findings should be done in separate issues

PR #1242, opened 2022-07-21
Update API npm packages

daniel-arnauer:
Update all npm packages to newest
Implement breaking changes

Closes  https://github.com/openkfw/TruBudget/issues/1089


frontend/src/pages/Analytics/ProjectAnalytics.js
```diff
@@ -213,18 +213,21 @@ const NumberChart = ({ title, budget, currency, dataTest }) => (
...
-      <Typography style={{ flex: 1 }} variant="overline">
-        {title}
-      </Typography>
-      <Typography style={{ flex: 1 }} data-test={dataTest} variant="h6">
-        {budget ? `${(budget * 100).toFixed(2)}%` : "-"}
-      </Typography>
-    </CardContent>
-  </Card>
-);
+const RatioChart = ({ title, budget, dataTest }) => {
```
Stezido, line 216: is this frontend change necessary for the api npm package updates?

Stezido (changes requested):

api/src/service/domain/workflow/workflowitem.ts
```diff
@@ -75,47 +75,46 @@ const schema = Joi.object().keys({
...
-    .when("status", {
-      is: Joi.valid("closed"),
-      then: Joi.required(),
-      otherwise: Joi.optional(),
+      otherwise: Joi.when("status", {
+        is: "closed",
+        then: Joi.required(),
+        otherwise: Joi.optional(),
+      }),
     }),
```
daniel-arnauer, line 98: not sure if this is correct

api/src/project_budget_delete_projected.ts
```diff
@@ -133,53 +133,55 @@ export function addHttpHandler(
   urlPrefix: string,
   service: Service,
 ) {
-  server.post(
-    `${urlPrefix}/project.budget.deleteProjected`,
-    mkSwaggerSchema(server),
-    (request, reply) => {
-      const ctx: Ctx = { requestId: request.id, source: "http" };
+  server.register(async function () {
```
daniel-arnauer, line 136: Maybe the server.register command could be used in index.ts to wrap over all Routes instead of wrapping every route on its own with server.register

frontend/src/pages/Analytics/ProjectAnalytics.js
daniel-arnauer: No but since it is a small fix in an extra commit, I see no problem

api/src/service/domain/workflow/workflowitem.ts
```diff
@@ -74,11 +74,11 @@ const schema = Joi.object().keys({
   createdAt: Joi.date().iso().required(),
   dueDate: Joi.date().iso().allow(""),
   displayName: Joi.string().required(),
-  // This should use exchangeRateSchema but can't, because of backward compatibility:
   exchangeRate: Joi.string()
     .when("amountType", {
       is: Joi.valid("N/A"),
       then: Joi.forbidden(),
+      break: true,
     })
```
daniel-arnauer, line 82: `break` was added to Joi in 16.x.x. This must be set to true, otherwise the next condition will be checked which is not the expected behavior.

mayrmartin (approved):

PR #1283, opened 2022-08-23
api: re-implement cache with node-cache to add ttl for cache entities

### LINKED ISSUE #1108: API heap problems
## Description 😯
The API runs out of heap in deployment. This is probably due to missing cache invalidation. There is a cache invalidation function but the function only gets executed on API restart.

There are several ways to solve this issue i.e. using a TTL or a Ring-Buffer or even a complete rewrite. If a complete rewrite is chosen as a solution 3rd party libraries ([also have a look at the data structures & their limitations](https://blog.appsignal.com/2021/12/08/nodejs-memory-limits-what-you-should-know.html))

**NB: This affects Trubudget 1.x and 2.0 so make sure to apply possible changes to both versions**

mayrmartin:
### Checklist

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/openkfw/TruBudget/blob/main/CONTRIBUTING.md).

### Description

Solves API-Heap problems by adding a TTL for cache entities

Closes #1108


api/src/service/cache/index.ts
```diff
@@ -1,168 +1,190 @@
...
-  cache.eventsByStream.clear();
-  cache.cachedProjects.clear();
-  cache.cachedSubprojects.clear();
-  cache.cachedWorkflowItems.clear();
-  cache.cachedSubprojectLookup.clear();
-  cache.cachedWorkflowitemLookup.clear();
-}
-
 interface CacheInstance {
+  //TODO: check which method must be async & which not
```
georgimld, line 103: please check if this todo is resolved and remove it once it is

georgimld:
This PR introduces a time to live for the cache - 10 mins. Every 10 mins the cache is cleared, after which a new request to the api (that uses the `withCache()`) function will trigger the cache to reload. Our purpose was to help with the api heap problems (issue #1108), however only a ttl does not fix the problem, because the cache is still filled again on a new request after it is cleared.

In order to fix the issue, we need to improve the caching startegy overall, which will be discussed in a separate discussion [#1290](https://github.com/openkfw/TruBudget/discussions/1290)-

PR #1308, opened 2023-01-24
bc: Increase request timeout of backup endpoints

Stezido:
### Description

This PR increases the timeout of the backup requests, download & restore.
If not increased large backups cannot be downloaded or restored.


Stezido:
for reviewer: pipeline can be ignored, since it's handling 2.x TruBudget only. The master branch is also deprecated.
The reason for the PR is a possibility for old versions to download/restore large backups.

georgimld (approved):

PR #1334, opened 2023-04-06
Validate assignee when creating workflow item

### LINKED ISSUE #1327: Workflow items with non-existing assignee cannot be modified

## Description 😯

During creating a workflow item directly through the API, when the value for the assignee is a user that does not exist in the system, workflow item is still created on-chain. However the workflow item cannot be modified at all because the "user not found" error is thrown.

### How to reproduce 🕹
1- Create a workflow item through API with assignee as a username that does not exist in the system. (e.g "exampleuser")
2- As a user with full permissions on the workflow item, try to modify the workflow item. (e.g changing exchange rate or bringing it to "paid" status)
3- Clicking on submit produces an error stating that the assigned user is not found and the request fails.

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| Frontend        | v2.1.0  |
| API             | v2.1.0  |
| Blockchain      | v2.1.0  |

egdmrsy:
### Checklist

### Description

2 validations on API Level are added:
1- Disabled users are not allowed to be assigned to workflow items.
2- Users that do not exist are not allowed to be assigned to workflow items.

Closes #1327

### DIFF
--- a/api/src/service/domain/workflow/workflowitem_create.spec.ts
+++ b/api/src/service/domain/workflow/workflowitem_create.spec.ts
@@ -41,6 +41,7 @@ describe("Create workflowitem", () => {
 
     const result = await WorkflowitemCreate.createWorkflowitem(ctx, root, data, {
       workflowitemExists: async (_projectId, _subprojectId, _workflowitemId) => false,
+      userExists: async (_userId) => false,
       getSubproject: async () => baseSubproject,
       applyWorkflowitemType: () => [],
       uploadDocumentToStorageService: () => Promise.resolve([]),
@@ -62,6 +63,7 @@ describe("Create workflowitem", () => {
 
     const result = await WorkflowitemCreate.createWorkflowitem(ctx, alice, data, {
       workflowitemExists: async (_projectId, _subprojectId, _workflowitemId) => false,
+      userExists: async (_userId) => false,
       getSubproject: async () => baseSubproject,
       applyWorkflowitemType: () => [],
       uploadDocumentToStorageService: () => Promise.resolve([]),
--- a/api/src/service/domain/workflow/workflowitem_create.spec.ts
+++ b/api/src/service/domain/workflow/workflowitem_create.spec.ts
@@ -41,7 +41,7 @@ describe("Create workflowitem", () => {
 
     const result = await WorkflowitemCreate.createWorkflowitem(ctx, root, data, {
       workflowitemExists: async (_projectId, _subprojectId, _workflowitemId) => false,
-      userExists: async (_userId) => false,
+      userExists: async (_userId) => true,
       getSubproject: async () => baseSubproject,
       applyWorkflowitemType: () => [],
       uploadDocumentToStorageService: () => Promise.resolve([]),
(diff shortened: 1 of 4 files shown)


api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -201,6 +204,17 @@ export async function createWorkflowitem(
     return new AlreadyExists(ctx, workflowitemCreated, workflowitemCreated.workflowitem.id);
   }
 
+  logger.trace({ item: workflowitemCreated }, "Check if assignee exists");
+  const isUserExists = await repository.userExists(
```
georgimld, line 208: in the TB api, we wrap return values in a `Result.Type`, which can also be an error - in this case you can check the [implementation of `userExists`](https://github.com/openkfw/TruBudget/blob/a0c42c2b74c532242eb3e87bda2e8324f622ecad/api/src/service/domain/organization/user_query.ts#L49) to see which errors it can return and in which cases.

Because of this, you must first check that the returned value is not an error and then use it the way you did. here is an example how this should be done: https://github.com/openkfw/TruBudget/blob/a0c42c2b74c532242eb3e87bda2e8324f622ecad/api/src/service/domain/organization/user_create.ts#L84

api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -201,6 +204,17 @@ export async function createWorkflowitem(
...
+    workflowitemCreated.workflowitem.assignee
+  );
+
+  if(!isUserExists) {
```
georgimld, line 212: I would also not name the const `isUserExists` - this sounds gramatically incorrect - just `userExists` is fine

api/src/service/domain/workflow/workflowitem_create.spec.ts
```diff
@@ -41,6 +41,7 @@ describe("Create workflowitem", () => {
 
     const result = await WorkflowitemCreate.createWorkflowitem(ctx, root, data, {
       workflowitemExists: async (_projectId, _subprojectId, _workflowitemId) => false,
+      userExists: async (_userId) => true,
```
georgimld, line 44: Please also add a new testcase that tests what you added, so if the user does not exist.

Also: what happens if you try this with a user that exists but is disabled? does this (and should this) still work?

api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -201,6 +207,42 @@ export async function createWorkflowitem(
...
+    workflowitemCreated.workflowitem.assignee
+    );
+
+  if (Result.isErr(userResult)) {
+    return new VError(userResult, "user check failed");
+  }
+
+  const userPermissions = userResult.permissions;
+
+  if(userPermissions["user.authenticate"] === undefined || userPermissions["user.authenticate"].length === 0) {
```
georgimld, line 238: the way you did it is fine but here's a short tipp how you can write cleaner code, you could refactor it this way:

`if(!userPermissions["user.authenticate"] || !userPermissions["user.authenticate"].length)`

api/src/service/domain/workflow/workflowitem_create.ts
```diff
@@ -201,6 +207,42 @@ export async function createWorkflowitem(
...
+    return new VError(userResult, "user check failed");
+  }
+
+  const userPermissions = userResult.permissions;
+
+  if(!userPermissions["user.authenticate"] || !userPermissions["user.authenticate"].length) {
+    return new PreconditionError(
+      ctx,
+      workflowitemCreated,
+      "disabled users are not allowed to create workflow items!"
```
georgimld, line 242: the message should be `disabled users are not allowed to be assigned to workflowitems`

georgimld (approved):

PR #1434, opened 2023-07-31
Snapshot logic

### LINKED ISSUE #1108: API heap problems

egdmrsy:
…s and workflows.

### Checklist

### Description

Closes #1108


api/src/service/workflowitem_cache_helper.ts
```diff
@@ -0,0 +1,56 @@
...
+  subprojectId: Subproject.Id,
+): Promise<Workflowitem.Workflowitem[]> {
+  const rpcClient = conn.multichainClient.getRpcClient();
+  let items: Item[] = [];
+  items = await rpcClient.invoke(
+    "liststreamkeyitems",
+    projectId,
+    subprojectId + "_workflows",
+    false,
+    0x7fffffff,
```
mayrmartin, line 38: comment this magic value or introduce a const

api/src/service/workflowitem_cache_helper.ts
```diff
@@ -0,0 +1,56 @@
...
+  );
+  let workflowitems: Array<Workflowitem.Workflowitem> = [];
+  let i;
+  for (i = 0; i <= items.length - 1; i++) {
+    if (items[i].data.json.type == "workflowitem_created") {
+      const result = await getWorkflowitem(
```
mayrmartin, line 44: consider using promise.allsettled then you could make this loop async maybe also using map/reduce functions

api/src/service/workflowitem_history_get.ts
```diff
@@ -24,20 +26,39 @@ export async function getWorkflowitemHistory(
...
+      getWorkflowitem: async (projectId, subprojectId, workflowitemId) => {
+        const rpcClient = conn.multichainClient.getRpcClient();
+        let items: Item[] = [];
+        try {
+          items = await rpcClient.invoke(
+            "liststreamkeyitems",
+            projectId,
+            workflowitemId,
+            false,
+            0x7fffffff,
```
mayrmartin, line 45: comment this magic value or introduce const

mayrmartin (changes requested):
Please add some documentation on how the caching works in the TruBudet Architecture documentation.

api/src/service/cache_snapshot.ts
```diff
@@ -0,0 +1,410 @@
...
+        return;
+      }
+      const parser = EVENT_PARSER_MAP[event.type];
+      if (parser === undefined) {
+        const eventType = event && event.type ? event.type : JSON.stringify(event);
+        logger.fatal({ streamName, item }, `Cache: Event type "${eventType}"" not implemented.`);
+        return;
+      }
+      return parser(event);
+    })
```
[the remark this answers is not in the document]
egdmrsy: Every item is mapped to have their parsed event in a new array and the events are returned. How would reducing everything down to a single item would be useful here?

api/src/service/subproject_cache_helper.ts
```diff
@@ -0,0 +1,43 @@
...
+  conn: ConnToken,
+  ctx: Ctx,
+  projectId: Project.Id,
+): Promise<Subproject.Subproject[]> {
+  const rpcClient = conn.multichainClient.getRpcClient();
+  let items: Item[] = [];
+  items = await rpcClient.invoke("liststreamkeyitems", projectId, "subprojects", false, 0x7fffffff);
+  let subprojects: Array<Subproject.Subproject> = [];
+  let i;
+  for (i = 0; i <= items.length - 1; i++) {
```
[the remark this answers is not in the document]
egdmrsy: same as above

api/src/service/cache_snapshot.ts
```diff
@@ -0,0 +1,413 @@
...
+import { VError } from "verror";
+import { Subproject } from "./domain/workflow/subproject";
+import { Project } from "./domain/workflow/project";
+import { Workflowitem } from "./domain/workflow/workflowitem";
+import getValidConfig from "../config";
+
+const STREAM_BLACKLIST = [
+  // The organization address is written directly (i.e., not as event):
+  "organization",
+];
```
mayrmartin, line 72: Local variable 'STREAM_BLACKLIST' is not used.

api/src/service/cache_snapshot.ts
```diff
@@ -0,0 +1,413 @@
...
+  return {
+    searchKey: searchKey,
+    sourceFromSnapshot: sourceFromSnapshot,
+    parseFromSnapshot: parseFromSnapshot,
+  };
+}
+
+async function getItemsInfo(
+  conn: ConnToken,
+  ctx: Ctx,
```
mayrmartin, line 308: Parameter 'ctx' of local function 'getItemsInfo' is not used.

api/src/service/cache_snapshot.ts
```diff
@@ -0,0 +1,413 @@
...
+        logger.warn(`ignoring event no item.data.json property found ${JSON.stringify(item)}`);
+        return;
+      }
+      if (event.intent) {
+        logger.warn(`ignoring event of intent ${event.intent}`);
+        return;
+      }
+      const parser = EVENT_PARSER_MAP[event.type];
+      if (parser === undefined) {
+        const eventType = event && event.type ? event.type : JSON.stringify(event);
```
mayrmartin, line 353: Condition 'event' is always true at this point because the false branch of the condition '!event' at line 343 has been taken.

mayrmartin (changes requested):

api/src/service/cache_snapshot.ts
```diff
@@ -0,0 +1,413 @@
...
+
+  // Traverse the list reverse to get latest snapshot
+  let snapshotIndex = -1;
+  for (let i = items.length - 1; i >= 0; i--) {
+    let item = items[i];
+    if (item.keys.includes(searchKey)) {
+      snapshotIndex = i;
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
egdmrsy: Counter-opinion: The snapshot events are usually near the end of the items array. That's why here we start looking from the end of the array to the beginning in reverse order, because for very long list of items this would be much more efficient and faster to find the snapshot. findIndex iterates in an ascending-index order. If we want to use findIndex, we have to in-place reverse the items array (redundant inefficiency for large items array) which messes up the event order, or we copy the array and reverse a copy of it but copying a whole array if the array size is big is not memory friendly.

api/src/service/subproject_cache_helper.ts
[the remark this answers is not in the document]
egdmrsy: For-loop is basic and makes the intend clear and code more readable. Also I might be wrong but as far as I know Array native functions like map and reduce are slower in comparison to naive for-loop approach, so that's why i think it would be overkill to use reduce for such simple example.

api/src/service/subproject_cache_helper.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export async function getAllSubprojects(
+  conn: ConnToken,
+  ctx: Ctx,
+  projectId: Project.Id,
+): Promise<Subproject.Subproject[]> {
+  const rpcClient = conn.multichainClient.getRpcClient();
+  let items: Item[] = [];
+  items = await rpcClient.invoke("liststreamkeyitems", projectId, "subprojects", false, 0x7fffffff);
```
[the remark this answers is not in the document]
mayrmartin: don't forget to use this constant you introduced 😃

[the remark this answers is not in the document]
egdmrsy: its already being used. the file here looks outdated

docs/developer/architecture/0013-snapshot-logic.md
```diff
@@ -0,0 +1,110 @@
...
+| subproject* |
+
+`workflowitem_snapshot_published`
+| Fields |
+--- |
+| projectId |
+| subprojectId |
+| workflowitem* |
+
+With * marked fields are aggregable objects that contain the latest event and item data relevant to the item type. Below you can find more info about how these objects are created and the worklow.
```
SamuelPull, line 64: typo: workflow

api/src/service/workflowitem_cache_helper.ts
```diff
@@ -0,0 +1,59 @@
...
+  return await SnapshotService.getLatestSnapshot(
+    ctx,
+    conn,
+    projectId,
+    workflowitemId,
+    "workflowitem_snapshot_published",
+  );
+}
+
+export async function getAllWorkflowitems(
```
SamuelPull, line 27: Opening any subproject results in a multichain rpc?

SamuelPull:
Are there any requirements for backward compatibility? Did we test if this could be an issue?

mayrmartin (approved):

PR #1772, opened 2024-04-15
storage-service:Possibility to use azure blob storage

### LINKED ISSUE #1656: Add additional interfaces for storage solutions
## Description 💡
At the moment the Storage Service can only connect and use MinIO as a storage backend. Other solutions which are based on the S3 protocol should be supported as well. Additionally, it should be possible to connect to other storage solutions with ease (i.e. Azure Blob Storage). MinIO provides a Gateway that can place files on different storage solutions using it would reduce the implementation/maintenance effort for us. Unfortunately, the gateway is deprecated but there are some alternatives (https://github.com/minio/minio/issues/14331)

galethil:
### Checklist

- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/openkfw/TruBudget/blob/main/.github/CONTRIBUTING.md#open-a-pull-request).

### Description
Implementation of Azure blob storage with local Azurite instance.

### How to test

1. Change env variable in storage service `STORAGE_PROVIDER=azure-blob`
2. Start local environment with `--use-azure-storage` e.g. `bash scripts/development/start-dev.sh --full --use-azure-storage --build`
3. Check if file functionality is working in app
4. ...

Closes #1656


SamuelPull:
STORAGE_PROVIDER has to be azure-storage, not azure-blob like in How to test.

I miss updated docs (environment-variables.md, architecture.md, developer-setup.md,...)

storage-service/src/azureblob.ts
```diff
@@ -0,0 +1,140 @@
+import { BlobServiceClient } from "@azure/storage-blob";
+import { v4 } from "uuid";
+import config from "./config";
+import { log } from "./index";
+import { FileWithMeta, Metadata, sleep } from "./storage";
+
+const containerName: string = config.azureBlobStorage.containerName;
+
+async function streamToString(readableStream): Promise<string> {
```
SamuelPull, line 9: Can be expressed as

async function streamToString(readableStream): Promise<string> {
  let data = "";
  for await (const chunk of readableStream) {
    data += chunk;
  }
  return data;
}

storage-service/src/azureblob.ts
```diff
@@ -0,0 +1,140 @@
...
+      chunks.push(data.toString());
+    });
+    readableStream.on("end", () => {
+      resolve(chunks.join(""));
+    });
+    readableStream.on("error", reject);
+  });
+}
+
+async function streamToBuffer(readableStream): Promise<Buffer> {
```
SamuelPull, line 22: If you think this is more readable you can use this

```
async function streamToBuffer(readableStream): Promise<Buffer> {
  const chunks: Buffer[] = [];
  for await (const chunk of readableStream) {
    chunks.push(chunk instanceof Buffer ? chunk : Buffer.from(chunk));
  }
  return Buffer.concat(chunks);
}
```

storage-service/src/config.ts
```diff
@@ -5,22 +5,43 @@ interface MinioConfig {
...
+  storageProvider: string;
   storage: MinioConfig;
+  azureBlobStorage: AzureBlobConfig;
 }
 
 const config: Config = {
   port: Number(process.env.STORAGE_SERVICE_PORT) || 8090,
   allowOrigin: process.env.ACCESS_CONTROL_ALLOW_ORIGIN || "*",
+  storageProvider:
+    process.env.STORAGE_PROVIDER === "azure-storage"
```
SamuelPull, line 26: Maybe just `azure` would be sufficient?

SamuelPull (approved):

storage-service/src/azureblob.ts
galethil: @SamuelPull I believe I copied this from official documentation. Your version is definitely nicer so I will use it.

storage-service/src/config.ts
galethil: @SamuelPull since azure is quite generic I wanted to be a little more specific. It is also the name of docker container running locally (since azure will be too generic for container) and I wanted it to be the same. Are you also ok with "azure-storage" or do you prefer to change it to "azure"?

SamuelPull: it's okay, thanks for the explanation

PR #1825, opened 2024-05-10
991 Default error handling - excel-export

galethil:
### Checklist

### Description

Add default error handling in excel export

### How to test

1. Start full environment
2. Simulate error in download route
3. Check if the http response is correct with necessary information
4. Check if the logs make sense and if they are descriptive.


excel-export-service/src/errors.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.name = this.constructor.name;
+    this.message = message;
+    this.status = status;
+    this.isPublic = isPublic;
+    this.originalMessage = originalMessage;
+  }
+}
+
+/**
+ * Executes callback function and if error occures, it sends a notification email.
```
SamuelPull, line 23: what notification email?  *confusion*

excel-export-service/src/index.ts
```diff
@@ -97,58 +100,91 @@ excelService.get("/version", (req: CustomExpressRequest, res: CustomExpressRespo
...
-
-  try {
-    await getApiVersion(axios, token, res.apiBase);
-  } catch (err) {
-    if (!err.response) {
-      logger.error({ err }, "Cannot connect to API service");
-      return res.status(503).send({ message: "Cannot connect to API service" });
+excelService.get(
+  "/download",
+  forwardError(async (req: CustomExpressRequest, res: CustomExpressResponse) => {
```
SamuelPull, line 105: Personally I don't like higher order function here when only one route is using it. I will aceept it, but I'd prefer 

excelService.get(
  "/download",
  async (req: CustomExpressRequest, res: CustomExpressResponse, next: NextFunction) => {
    try {
      ...
     } catch (err) {
      next(err);
    }
  },
);

It's up to you.

excel-export-service/src/errors.ts
```diff
@@ -0,0 +1,39 @@
+import { NextFunction } from "express";
+import { CustomExpressRequest, CustomExpressResponse } from "./types";
+
+/**
+ * * @extends Error
+ */
+export class ExtendableError extends Error {
+  status: number;
```
SamuelPull, line 8: If status means http status code, shouldn't it be a member of APIError?

excel-export-service/src/apiError.ts
```diff
@@ -0,0 +1,15 @@
...
+import { ExtendableError } from "./errors";
+
+/** * Class representing an API error. *
+ * @extends ExtendableError
+ * @param {string} message - Error message.
+ * @param {number} status - HTTP status code of error.
+ * @param {boolean} isPublic - Whether the message should be visible to user or not.
+ * @param {string | null} originalMessage - Original Error message that was thrown.
+ */
+export class APIError extends ExtendableError {
```
SamuelPull, line 11: Q: Why not extend Error directly?

excel-export-service/src/index.ts
```diff
@@ -97,58 +100,91 @@ excelService.get("/version", (req: CustomExpressRequest, res: CustomExpressRespo
...
+    } catch (error) {
+      req.log.error({ err: error }, "Error while creating excel export");
+      if (error.response) {
+        return res.status(error.response.status).send({ message: error.response.data });
+      }
     }
+  }),
+);
+
+excelService.use(function (
```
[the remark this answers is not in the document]
galethil: @SamuelPull this is the official way of express default error handling described in documentation. 
image

But I will add name and comment to make it more straight forward.

excel-export-service/src/errors.ts
galethil: @SamuelPull Sorry, my bad. I used the structure from my different project where we have email notifications set up on production errors. I'm changing the comment.

excel-export-service/src/index.ts
galethil: @SamuelPull I'm open for discussion how to do it. Since this should be something implemented also in other services. For me there are 2 advantages to this approach ones you addopt it.
- It is much quicker with less code. You don't need to duplicate try/catch block in every route
- And you can easily extent the code `forwardError` function in case you need better error handling or any additional functionality.
If you have better way to do it. I'm open to other options.

excel-export-service/src/apiError.ts
```diff
@@ -0,0 +1,15 @@
+import httpStatus = require("http-status");
```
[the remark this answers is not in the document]
galethil: @SamuelPull in the past I found using of this library usefull in some cases. But it is not neccesary to use it. I removed it.

excel-export-service/src/apiError.ts
galethil: @SamuelPull It might be useful for more granular error management. E.g. if app is communicating not just via Rest api, but also sockets or Kafka messages, which is not the case in this instance, so I removed the ExtendableError to simplify the structure.

excel-export-service/src/errors.ts
```diff
@@ -0,0 +1,39 @@
...
+  isPublic: boolean;
+  originalMessage: string | null;
+
+  constructor(message: string, status: number, isPublic: boolean, originalMessage: string | null) {
+    super(message);
+    this.name = this.constructor.name;
+    this.message = message;
+    this.status = status;
+    this.isPublic = isPublic;
+    this.originalMessage = originalMessage;
```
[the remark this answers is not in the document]
galethil: @SamuelPull Again might be useful if you produce error message which you don't want to be public, but you still want to keep it as it travel throw the layers of application. For now I removed it.

excel-export-service/src/errors.ts
galethil: ExtendableError removed

SamuelPull (approved):

excel-export-service/src/apiError.ts
SamuelPull: It's hard to always distinguish between a thoughtful anticipation of changing requirements, and littering the code with unnecessary design elements. Maybe the future will tell that this class hierarchy was a better choice.

PR #1827, opened 2024-05-10
first version of discussion guide for the IT department

jzakotnik:
Update of the TruBudget documentation based on recent discussions with IT departments. The markdown should provide an initial view what to think about.

If I missed anything, please comment here and I can update the relevant items..


SamuelPull:
Perhaps how to configure `nginx` server for frontend?

jzakotnik:
Thanks! I added a section about DMZ... the idea is to have a small list of questions for non-IT project managers to have a first impression with the IT department.. so nginx is already pretty technical and hopefully covered in the install guide.

SamuelPull (approved):

PR #1910, opened 2024-07-08
frontend: Call authproxy signout on logout

### LINKED ISSUE #1855: User directory adapter: Login possible even though permissions/access was revoked
- Related to https://github.com/openkfw/TruBudget/issues/1648 (user directory adapter)
### How to reproduce 🕹

1. Login using the user directory adapter
2. Remove the permissions/access in the user directory
3. Login again using the user directory adapter

Expected behavior: login is denied as the permission/access is missing
Actual behavior: login is possible, as the browser uses cached responses from previous logins, which were successful
Workaround: delete cookies/cache - but might not be suitable

## Your Environment 🌎

| Tech            | Version |
| --------------- | ------- |
| Frontend        | v2.10.0  |
| API             | v2.10.0  |
| Blockchain      | v2.10.0  |
| Provisioning | v2.10.0 |
| Export-Service  | N/A |
| Email-Service   | N/A |
| Storage-Service | N/A  |
| Authbuddy | 1.1.0 |

SamuelPull:
### Checklist

### How to test

1. If testing locally, start an Auth Proxy instance
3. Start TB configured to use Auth Proxy (AUTHPROXY_ENABLED, AUTHPROXY_JWS_SIGNATURE)
2. Login with Auth Proxy
3. Logout 
4. You should be logged out of TB, and another window should open to sign out from Auth Provider
5. If logging in normally, no additional window should pop up

Closes #1855


frontend/src/api.js
```diff
@@ -5,11 +5,18 @@ import _isEmpty from "lodash/isEmpty";
...
 import { base64ToBlob } from "./helper";
 import strings from "./localizeStrings";
+import { store } from "./store";
 
 const devMode = config.envMode === "development";
 const API_VERSION = "1.0";
 const instance = axios.create();
 
+const isAuthProxyEnabled = window?.injectedEnv?.REACT_APP_AUTHPROXY_ENABLED === "true" || config.authProxy.enabled;
+const authProxySignoutUri = (window?.injectedEnv?.REACT_APP_AUTHPROXY_URL || config.authProxy.url).replace(
```
galethil, line 15: @SamuelPull this feels a little strange. I would probably prefer separate variable for signout or and appendix to the original URL, but we can keep it like this.

frontend/src/api.js
```diff
@@ -91,7 +98,15 @@ class Api {
       token
     });
 
-  logout = () => instance.post(`/user.logout`, {});
+  logout = () => {
+    if (isAuthProxyEnabled) {
+      const isUsingAuthproxy = store?.getState()?.get("login")?.toJS()?.isUsingAuthproxy;
+      if (isUsingAuthproxy && authProxySignoutUri) {
+        window.open(`${authProxySignoutUri}`, "_blank");
```
galethil, line 105: @SamuelPull add detection of blocked popup and at maybe just display a message to the user? https://stackoverflow.com/a/2917/3198533

SamuelPull: @galethil Thank you, I was looking for a way how to handle this potential issue.
For testing purposes, I executed `window.open = function() { };`, because I could not block popups on localhost.

frontend/src/api.js
SamuelPull: Yeah, it would have been better to use SOME_URL and append to it, but that's a breaking change now. Let's keep it like this. Not great, not terrible.

galethil (approved):

PR #1946, opened 2024-08-20
1906-show all tags

### LINKED ISSUE #1906: Only one tag is shown in the workflow item overview even if more are present
We added various tags to a workflow item, but only the first is rendered. Seems in the code only 1 tag is used?

```
<div className="tag-cell">
                  {tags.length > 0 && (
                    <Chip
                      label={tags[0]}
                      size="small"
                      onClick={(event) => {
                        props.storeWorkflowitemSearchTerm(`tag:${event.target.innerText}`);
                      }}
                      sx={{ backgroundColor: (theme) => theme.palette.tag.main, color: "white" }}
                    />
                  )}
                </div>
```

image

MartinJurcoGlina:
### Checklist

- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/openkfw/TruBudget/blob/main/.github/CONTRIBUTING.md#open-a-pull-request).

### Description

* added popover that shows the rest of the tags in workflow items

### How to test

1. log into the app and open any project and its subproject
2. add tags to any workflow item where possible and after submission check outcome on workflow item page
3. try different possibilities like: add only one tag - expand button should not be present 
4. try adding several tags - expand button should be present besides the tag
5. after clicking expand button, all tags should be visible in popover

Closes #1906


SamuelPull (approved):

SamuelPull (changes requested):
Some e2e tests are failing, because the elements don't fit in the default cypress viewport size (1000x660).


I'll update with more info

SamuelPull (approved):

